PR #18, opened 2024-05-28
Detekt

### LINKED ISSUE #6: Introduce Detekt
### Problem

Ensuring consistency in coding style can easily become challenging. Without a standardized approach, the codebase may accumulate technical debt, leading to reduced maintainability.

### Solution

Introducing [detekt](https://detekt.dev/). Detekt is a static code analysis tool specifically designed for Kotlin, which is able to detect code smells, and enforces coding style conventions. It can be setup locally and integrated into CI/CD pipelines. We can therefore automate code quality checks and enforce best practices consistently across the codebase

### Additional context

_No response_

### Code of Conduct

eldcn:
# Add Detekt

Fixes #6

- Adds Detekt in the project
- Configures github action on pull requests
- Add pre-commit detekt git hook
- Add internal guide to setup detekt plugin locally and pre-commit hook installation

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: Detekt review
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4.1.0
+        with:
+          fetch-depth: 1
+          ref: ${{ github.head_ref }}
+
+      - name: Detekt PR Check
```
eldcn, line 22: @PSchmiedmayer This is the detekt action that uses [reviewdog](https://github.com/marketplace/actions/run-detekt-with-reviewdog)

Regardless of the linter result, `Detekt review` job will always succeed even if it finds any issues.  There are two options to configure for `reviewdog_reporter: (github-pr-check, github-pr-review)` and I am using `github-pr-check` github-pr-check reporter reports results to [GitHub Checks](https://github.com/reviewdog/reviewdog#reporter-github-checks--reportergithub-pr-check).

Detekt gradle task is also included in the `build` task of the project, hence if there are any issues `Build and test` job fails as expected due to detekt issues. Ideally, we shouldn't proceed with `Build and test` if we already seen issues in the `Detekt review` job. I tried multiple configurations but nothing worked out so far, left both jobs to run in parallel for now. I have intentionally included several warnings in `CoroutinesModule.kt` file in this PR for testing purposes.

core/coroutines/src/main/kotlin/edu/stanford/spezi/core/coroutines/di/CoroutinesModule.kt
eldcn, line 1: @PSchmiedmayer file that contains violations

eldcn: Update: I reverted the violations so that CI passes, in order to reproduce the issue feel free to push a violation in any of the kotlin files of the project, e.g. multiple empty lines between functions

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/Contact.kt
```diff
@@ -27,7 +27,7 @@ data class Contact(
     val description: String,
     val organization: String,
     val address: String,
-    val options: List<ContactOption>
+    val options: List<ContactOption>,
```
eldcn, line 30: See Kotlin's [trailing commans convention](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas)

.github/workflows/merge.yml
```diff
@@ -0,0 +1,55 @@
...
+  push:
+    branches: [ "main" ]
+
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
PSchmiedmayer, line 13: Just to make it clearer what this job does it might make sense to name them accordingly, e.g.,:
```suggestion
  buildandunittest:
    runs-on: ubuntu-latest
    name: Build and Run Unit Tests
```

.github/workflows/merge.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: gradle
+
+      - name: Build with Gradle
+        run: chmod +x ./gradlew && ./gradlew build
```
PSchmiedmayer, line 26: As far as my understanding goes build still invokes all unit tests across all modules/core/app elements of the projects.

As far as I can see it doesn't execute the Android/UI tests. Does it make sense to have a separate action that executes the Android/UI tests?
I suspect it's the same setup but just `./gradlew connectedAndroidTest`? Does it make sense to add this as a step after the first one right in this job or should it be a separate job? (I am leaning more to a separate job to parallelize this, on the other hand it might not be needed for the current scale ...).

.github/workflows/pull_request.yml
PSchmiedmayer, line 1: The same/similiar comments as in the merge.yml file also apply to this GitHub action.

.github/workflows/merge.yml
```diff
@@ -0,0 +1,55 @@
...
+on:
+  push:
+    branches: [ "main" ]
+
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
```
PSchmiedmayer, line 12: I added a few comments here, but they are generally also applicable to the main pull request GitHub action. 

It might even make sense to ensure that we don't duplicate this by just calling the PR GitHub action in its completeness and move almost everything except for the documentation deployment over there? 

E.g., we do a similar thing in our Spezi iOS Repos: https://github.com/StanfordSpezi/SpeziTemplateApplication/blob/fafb5c6c30f9e049b6515e7ea9f46af624c309a1/.github/workflows/beta-deployment.yml#L18

We could simply do something like
```yml
buildandtest:
    name: Build and Test
    uses: ./.github/workflows/pull_request.yml
    permissions:
      security-events: write
      packages: read
      actions: read
      contents: read
    secrets: inherit
```

.github/workflows/merge.yml
PSchmiedmayer, line 1: Small styling question: I would suggest to name this action `main` to be consistent with some of our other repos.

.github/workflows/pull_request.yml
PSchmiedmayer, line 1: Small styling question: I would suggest to name this action `build-and-test.yml` to be consistent with some of our other repos, especially when we reference it from the `main` action.

PSchmiedmayer (approved):
Thank you for all the work here @eldcn!

I had a few suggestions, especially focusing on the CI setup and elements around there that might be helpful.

Once these are addressed feel free to merge the PR 🚀

Thanks for all the work to get the infrastructure in place!

.github/workflows/merge.yml
eldcn: > As far as my understanding goes build still invokes all unit tests across all modules/core/app elements of the projects.

Yes, that's correct 👍 

I pointed it out in our last Android meeting that we are not executing UI tests, and I would include it next in separate PR. Reason is that it does not simply involve running `./gradlew connectedAndroidTest`, but it requires additional setup beforehand, such e.g. setting up an emulator where the ui tests will run. Haven't ever done it with Github actions before, hence would like to address it separately, because I suspect I might need some commits to get it running correctly 😅

PSchmiedmayer: Sounds good! Let's tackle this in a separate PR 👍

.github/workflows/pull_request.yml
eldcn: Please have a look at this pipeline: https://github.com/StanfordSpezi/SpeziKt/actions/runs/9309515966/job/25625090503

Even though we define just two jobs, the detekt action running in `Detekt review` job, additionally creates a third `detekt` job. This is the job that fails actually if there are some violations. Ideally here we should stop the whole pipeline if that's the case, however I did not succeed to do that, even when specifying a fail fast strategy.

> I am wondering if we could simply run `./gradlew detekt` right after this action as part of this job as a separate run step after this one?

That is not necessary in my opinion. The detekt github action is way more lightweight in comparison to  `./gradlew detekt` gradle task. The github action operates on the source code directly rather than on artifacts and does not require a project build. It executes the packaged JAR java application: `java -jar /opt/detekt.jar --config internal/detekt-config.yml --report xml:detekt_report.xml --excludes **/build/**,**/.idea/** --plugins /opt/detekt-formatting.jar`.

On the other hand, `./gradlew detekt` would require a project build beforehand in order to run. We could theoretically do that, but it would be redundant in my opinion as that detekt gradle task is already part of the `build` (`check`) gradle task (See [here](https://detekt.dev/docs/gettingstarted/gradle/#disabling-detekt-from-the-check-task)). We can exclude as proposed in the link and we would need to explicitly run `./gradlew detekt` after `build`

I think if everyone will follow the guide with setting up detekt locally (Android studio plugin and the git hook), it will ideally never be the case that the detekt job will fail.

.github/workflows/merge.yml
eldcn: Created this issue https://github.com/StanfordSpezi/SpeziKt/issues/21 and suggested in addition to implement a screen recording mechanism 👍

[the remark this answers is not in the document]
eldcn: worked like a charm 🚀 thank you 🙏

.github/workflows/merge.yml
eldcn: liked this a lot 🚀 TIL moment 💯 

Applied and will observe after merging

Issue #42, opened 2024-06-14
Local Storage Encryption and Key Management


PSchmiedmayer:
### Problem

Similar to [Spezi Storage](https://github.com/StanfordSpezi/SpeziStorage) we should have a nice reusable and ideally even more abstracted API than the Google Jetpack Security API to facility the storage of encrypted files storing protected health information. 

### Solution

We can utilize and build a small reusable wrapper around https://developer.android.com/reference/androidx/security/crypto/EncryptedFile to ensure that people use these APIs following best practices and pointing people the modern security APIs on Android.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #47, opened 2024-06-16
Feature/issue 42 local storage encryption

Basler182:
# *Feature/issue 42 local storage encryption*

#42 

## :gear: Release Notes 

- Added Key-Value Storage (EncryptedFileStorage)
- File Storage (EncryptedSharedPreferencesStorage as well as LocalStorage)


PSchmiedmayer (commented):
Thank you @Basler182 for working on this; this will be helpful for ENGAGE-HF and a lot of future projects!

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
```diff
@@ -0,0 +1,57 @@
...
+    private fun getEncryptedFile(fileName: String): EncryptedFile {
+        val file = File(context.filesDir, fileName)
+
+        return EncryptedFile.Builder(
+            context,
+            file,
+            masterKey,
+            EncryptedFile.FileEncryptionScheme.AES256_GCM_HKDF_4KB
+        ).build()
+    }
```
eldcn, line 31: Consider my proposal comment with the `FileHandler` component.

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    override suspend fun deleteFile(fileName: String) = withContext(ioDispatcher) {
+        val file = File(context.filesDir, fileName)
+        if (file.exists()) {
+            file.delete()
+        }
+    }
+}
```
eldcn, line 57: File operations are `throwing`, we should consider error handling throughout this component (mainly `runCatching` wrappings

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
```diff
@@ -0,0 +1,57 @@
...
+import dagger.hilt.android.qualifiers.ApplicationContext
+import edu.stanford.spezi.core.coroutines.di.Dispatching
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.withContext
+import java.io.File
+import javax.inject.Inject
+
+class EncryptedFileStorage @Inject constructor(
+    @ApplicationContext private val context: Context,
+    @Dispatching.IO private val ioDispatcher: CoroutineDispatcher,
```
eldcn, line 15: 💪

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/LocalKeyValueStorage.kt
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    override fun <T> readData(key: PreferenceKey<T>): Flow<T?> {
+        return dataStore.data.map { preferences ->
+            preferences[key.key]
+        }
+    }
+
+    override suspend fun <T> readDataBlocking(key: PreferenceKey<T>): T? {
+        return dataStore.data.first()[key.key]
```
eldcn, line 38: Prefer using `firstOrNull` as `first` will throw if the flow is empty

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/LocalKeyValueStorage.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val dataStore = context.dataStore
+
+    override suspend fun <T : Any> saveData(key: PreferenceKey<T>, data: T) {
+        dataStore.edit { preferences ->
+            preferences[key.key] = data
+        }
+    }
+
+    override fun <T> readData(key: PreferenceKey<T>): Flow<T?> {
+        return dataStore.data.map { preferences ->
```
eldcn, line 32: In general all `dataStore.data` should be followed by `dataStore.data..catch { emit(emptyPreferences()) }` to avoid throwing and using an empty preferences file in case something went wrong when accessing the file

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorage.kt
```diff
@@ -0,0 +1,10 @@
+package edu.stanford.spezi.modules.storage.key
+
+import kotlinx.coroutines.flow.Flow
+
+interface KeyValueStorage {
```
eldcn, line 5: I find it great that you added storage support with both data store and shared preferences API 💪 I am noting down some suggestions regarding the API of this module.

1. I would start with a singleton Factory component `KeyValueStorageFactory` with a method, `createStorage(name: String, type: (DataStore/SharePrefs): KeyValueStorage` The factory itself will hold the created storages in a HashMap, if already a storage of the same name and type exists, return that, otherwise create new. It is important to not create more than one data store with the same name, hence the map is important. The `LocalKeyValueStorage` and `EncryptedSharedPreferencesKeyValueStorage` can be marked internal and no need for `@Inject` constructor, since it will be created inside the factory. Extension property datastore on the companion object can move in the class level.

Why a factory? Well, key value storages keep simple values or settings which might also be clustered by feature, hence it makes sense to support creation of different files. The factory will prefix the filenames with `edu.stanford.spezi`, e.g. `APP_SETTINGS` filename will result into `edu.stanford.spezi.APP_SETTINGS`. This allows having all files sorted / grouped in device explorer.

2. Instead of `PreferenceKey<T>` sealed class, I would propose explicit methods for supported types, e.g. readInt, putInt. In the LocalKeyValueStorage implementation, you can use `intPreferencesKey(key)` as an implementation detail, and the exposed `key` of all method signatures will be of type `String`:

```kotlin
suspend fun putInt(key: String, value: Int) {
    dataStore.edit { preferences ->
        preferences[intPreferencesKey(key)] = value
    }
}

suspend fun readInt(key: String): Int? {
    return preferences()
        .map { preferences ->
            preferences[intPreferencesKey(key)]
        }.firstOrNull()
}

private fun preferences() = dataStore.data.catch { emit(emptyPreferences()) }
```

As writing and reading to storages is a pretty common use cases, I think having explicit methods simplifies the API in comparsion to the generic PreferenceKey. We can then create provide default hilt dependencies with a given file name for data store and shared preferences implementation, also enable creation of any arbitrary file name via the factory.

modules/storage/README.MD
```diff
@@ -0,0 +1,80 @@
...
+The storage module provides components for managing storage in your application. It includes classes
+for handling key-value/file storage and secure storage of data.
+
+## Usage
+
+To use the Storage module in your project, add the following dependency to your `build.gradle` file:
+
+```gradle
+dependencies {
+    implementation(":core:storage")`
```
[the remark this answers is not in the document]
Basler182: Well seen 👍

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/LocalKeyValueStorage.kt
Basler182: Good Idea :)

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorage.kt
Basler182: Thanks for the feedback. 

I like the first point as it offers more flexibility. 

I am unsure about the second point. The _SharedPreferences_ Library uses an approach that I don't particularly like. I therefore used the more advanced approach of the _DataStore_ implementation and took this as inspiration for the interface which is used by both implementations. I have adapted it so that the _SharedPreferences_ implementation also uses the same interface. I implemented missing features, such as _ByteArray_, custom so that the respective implementations can be exchanged if necessary.

With the data store implementation you can simply create objects like:
`private object PreferencesKeys {
  val SHOW_COMPLETED = booleanPreferencesKey("show_completed")
  val USER_NAME = stringPreferencesKey("user_name") 
}`

and than use the generic method to retrieve any of them. 

So if you have a list of preferences, which I think can happen quickly, you can just dynamically load them all out here, whereas if you have a separate method for each data type, you always have to pick the right method or you would probably write your own class that abstracts this logic, which would actually bring you back to this approach, right?

modules/storage/src/androidTest/kotlin/edu/stanford/spezi/modules/storage/key/EncryptedSharedPreferencesKeyValueStorageTest.kt
```diff
@@ -0,0 +1,171 @@
+package edu.stanford.spezi.modules.storage.key
```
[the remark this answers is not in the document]
Basler182: Exactly, I had exactly the same problem with the missing application context that you described. I also came across Robolectric, but initially decided to check the entire functionality with an Android test. At that point, I also thought it made sense to let the user configure as little as possible in the part where encryption is important to avoid possible errors. However, with the factory method you mentioned in another comment, we could work around this issue, so it probably make sense to customise the implementation based on your suggestion. 👍

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
Basler182: Good Idea! 👍

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/PreferenceKey.kt
```diff
@@ -0,0 +1,20 @@
...
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.booleanPreferencesKey
+import androidx.datastore.preferences.core.byteArrayPreferencesKey
+import androidx.datastore.preferences.core.doublePreferencesKey
+import androidx.datastore.preferences.core.floatPreferencesKey
+import androidx.datastore.preferences.core.intPreferencesKey
+import androidx.datastore.preferences.core.longPreferencesKey
+import androidx.datastore.preferences.core.stringPreferencesKey
+
+sealed class PreferenceKey<T>(val key: Preferences.Key<T>) {
```
eldcn, line 12: Let's mark key as `internal val key`, otherwise consumers of this module would need to add datastore dependency

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    override suspend fun deleteFile(fileName: String): Result<Unit> = withContext(ioDispatcher) {
+        runCatching {
+            val file = File(context.filesDir, fileName)
+            if (file.exists()) {
+                file.delete()
+            }
+        }
+    }
+}
```
eldcn, line 65: I think we can get rid of `Result<Unit>` return types in save and delete methods, nothing we can do to recover in case something goes wrong, `runCatching { // impl }.onFailure { // log error }` should be enough

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
```diff
@@ -0,0 +1,65 @@
...
+            runCatching {
+                deleteFile(fileName)
+                val encryptedFile = getEncryptedFile(fileName)
+                encryptedFile.openFileOutput().use { outputStream ->
+                    outputStream.write(data)
+                }
+            }
+        }
+
+    override suspend fun readFile(fileName: String): Result<ByteArray?> =
```
eldcn, line 44: Also here, `: ByteArray?` return type should be enough: `runCatching { }.onFailure { // log error }.getOrNull()`

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorage.kt
eldcn: I see 👍 I am already in favor of supporting both APIs as I mentioned. Even though shared preferences is old, it is still massively used from tons of projects due to its simple and reliable API. You can have a look at this [article](https://medium.com/mobile-innovation-network/sharedpreferences-vs-datastore-choosing-the-right-one-72f0789c956c#:~:text=When%20deciding%20between%20Jetpack%20DataStore,DataStore%20is%20a%20better%20choice.) that summarizes the diffs. From my perspective, considering our use cases we should stick with SharedPrefs only. For more complex types, we could persist json string of the instance. This would mean, we support read/write for primitive types only (no need for byte array), and add extension functions on `KeyValueStorage` for read/write serializable at a given key, which internally will encode / decode to json string. No matter what, it is fine as it is now 👍 

My second point was mainly about API design:

```kotlin
suspend fun handle(storage: LocalKeyValueStorage) {
    val key = "someKey"
    // 1. Option
    val value1 = storage.readDataBlocking(key = PreferenceKey.IntKey(name = key))
    // 2. Option
    val value2 = storage.readInt(key = key)
}
```

I personally find option 2 more lightweight and less boilerplate 😅 However, both options are fine by me 👍

eldcn (commented):
LGTM 🚀  Are you planning to adapt factory approach with support for multiple storage files as part of this PR?

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorage.kt
Basler182: Adding json later seems like a good idea.

at least my idea would have been that people use it more like this:

`private object UserData { 
val SHOW_COMPLETED = PreferenceKey.BooleanKey("show_completed") 
val USER_NAME = PreferenceKey.StringKey("user_name")
}`

and than it can be read like:

`keyValueStorage.readDataBlocking(SHOW_COMPLETED)
keyValueStorage.readDataBlocking(USER_NAME)
`

modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/file/EncryptedFileStorage.kt
Basler182: am i missing something or is this really the behaviour you would want? especially if i save something and it doesn't work, it would be helpful to know that the result failed so that i can react to it and e.g. try again later depending on the error, or save it in a different way, right ?

eldcn (approved):

PR #16, opened 2024-05-25
Optimize GitHub Actions Workflow for Android CI

Basler182:
# *Optimize GitHub Actions Workflow for Android CI*

## :recycle: Current situation & Problem
*Link any open issues or pull requests (PRs) related to this PR. Please ensure that all non-trivial PRs are first tracked and discussed in an existing GitHub issue or discussion.*

## :books: Documentation

- add google-services.json in workflow 
- allow manual triggering of workflow

## :pencil: Code of Conduct & Contributing Guidelines 

By submitting creating this pull request, you agree to follow our [Code of Conduct](https://github.com/StanfordSpezi/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/StanfordSpezi/.github/blob/main/CONTRIBUTING.md):
- [x] I agree to follow the [Code of Conduct](https://github.com/StanfordSpezi/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/StanfordSpezi/.github/blob/main/CONTRIBUTING.md).

PR #30, opened 2024-06-09
Fastlane Setup & Google Play Alpha Deployment

PSchmiedmayer:
# Fastlane Setup & Google Play Alpha Deployment

## :recycle: Current situation & Problem
- The GitHub Actions currently only create an apk that is not downloadable from the Google Play store 

## :gear: Release Notes 
- Adds an initial fastlane setup
- Adds GitHub Action to publish to the play store
- Adds content metadata about the ENGAGE-HF app

## :pencil: Code of Conduct & Contributing Guidelines 

By submitting creating this pull request, you agree to follow our [Code of Conduct](https://github.com/StanfordSpezi/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/StanfordSpezi/.github/blob/main/CONTRIBUTING.md):
- [x] I agree to follow the [Code of Conduct](https://github.com/StanfordSpezi/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/StanfordSpezi/.github/blob/main/CONTRIBUTING.md).

PR #48, opened 2024-06-19
Setup Android Tests CI Setup

### LINKED ISSUE #21: Execute `connectedAndroidTest` in CI
### Problem

Currently we are not executing any UI tests in the pipeline, but only the unit tests as part of the `build` gradle tasks. This way we might miss issues in our UI tests, as they need to be executed explicitly locally.

### Solution

- Include execution of UI tests as part of the pull requests workflow
- Implement a screen recording mechanism of running UI tests (via a test rule), and upload video files only for failing tests. This way we can analyze directly in the pipeline what when wrong

### Additional context

_No response_

### Code of Conduct

PSchmiedmayer:
# Setup Android Tests CI Setup

- Addresses #16
- Closes #21

## :gear: Release Notes 
- Setup Android Tests CI Setup

## ☑️ Remaining Issues
- [x] Report Code Coverage: I have build the rough setup for running the tests but the code coverage report currently doesn't work as expected for the Android tests. Codecov seems to merge reports for the same commit so we can just upload them all in different builds. Would be amazing if you can take a look at this @eldcn 
- [x] Contact Tests Fail: It seems like the contact Android tests fail locally and on the CI. Good as a test for the setup, would be great if we can fix them as part of this PR @Basler182 


fastlane/Fastfile
```diff
@@ -9,11 +9,16 @@
 default_platform(:android)
 
 platform :android do
-  desc "Runs all the tests"
+  desc "Runs all unit tests"
   lane :test do
     gradle(task: "test")
   end
 
+  desc "Runs all UI tests"
```
eldcn, line 17: As we pass the script to the github action, this is currently not used, can we remove it? I tried to use but the bundle wasn't available in the action's environemnt

gradle/libs.versions.toml
```diff
@@ -2,12 +2,12 @@
 [versions]
 accompanistPager = "0.35.1-alpha"
 activityCompose = "1.9.0"
-agp = "8.4.1"
-androidTools = "31.4.2"
+agp = "8.5.0"
```
eldcn, line 5: 💪 

Note for everyone: @vishnuravi @Basler182, please make sure to update your android studio to Koala after this PR will get merged (see [Android Gradle plugin and Android Studio compatibility](https://developer.android.com/build/releases/gradle-plugin#android_gradle_plugin_and_android_studio_compatibility))

build.gradle.kts
```diff
@@ -111,17 +111,36 @@ fun Project.setupJacoco() {
...
+         * We enable androidTests coverage for debug builds in SpeziBaseConfigConventionPlugin.
+         * There is however an issue when generating jacoco report of unit and android tests
+         * as the classes are already instrumented from the former test type. We follow the solution
+         * from this SO answer: https://stackoverflow.com/a/68481159
+         * We use release execution data for unit test coverage, and debug data for androidTests
+         */
+        executionData.setFrom(
+            files(
+                "$buildDir/outputs/unit_test_code_coverage/releaseUnitTest/releaseDebugUnitTest.exec",
+                "$buildDir/outputs/code_coverage/debugAndroidTest/connected/*.ec"
```
eldcn, line 138: The initial issue that you faced, was because the `connected.ec` file, which jacoco uses to create the code coverage reports was not being created. For that it is needed that we explicitly enable `enableAndroidTestCoverage` (added in `SpeziBaseConfigConventionPlugin`). Then I faced the issue in the comment and followed the solution of linked answer.

.github/workflows/build-test-analyze.yml
```diff
@@ -69,6 +69,71 @@ jobs:
           flags: unittests
           name: codecov-coverage
           token: ${{ secrets.CODECOV_TOKEN }}
+  test:
```
eldcn, line 72: Since we already run unit tests and upload the unit tests coverage as part of `buildtestandanalyze` (which runs on pull requests), would it make sense to merge the steps of a single matrix configuration in `buildtestandanalyze` and upload coverage once? We can then think of executing all matrix configurations of this current job when merging to main. Also, I noticed that there have been [updates](https://github.com/StanfordSpezi/SpeziKt/pull/30) in the workflows, we are now also deploying Dokka documentation on every PR. I missed the review from #30, should we move `dokka` job back to main workflow?

.github/workflows/build-test-analyze.yml
```diff
@@ -69,6 +69,71 @@ jobs:
...
+    name: Instrumented tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
eldcn, line 76: I think we can fail fast here, `connectedCheck` is a relatively long lasting task 😅

eldcn (approved):

Basler182 (approved):
Looks great to me. Thanks, @PSchmiedmayer, for creating it and @eldcn for fixing the bug in the test and updating the dependencies.

fastlane/Fastfile
PSchmiedmayer: Agree; I will remove it 👍

.github/workflows/build-test-analyze.yml
PSchmiedmayer: Good point; I had it in there to do some testing but we should enable it from now on 👍

build.gradle.kts
PSchmiedmayer: Amazing; thank you @eldcn; this looks great and I looked at the diff to learn what you have changed, very nice! 👍

.github/workflows/build-test-analyze.yml
PSchmiedmayer: I tried to look into making this a single matrix but it gets tricky with the fact that we need a different setup for unit tests and Android tests + we only really need variations on the devices + SDK levels for the connected tests. I would therefore suggest to keep it that way for now, the reference from the main branch will execute:
1. All unit tests + upload the code coverage there
2. All connected tests in a test matrix of devices + android versions that we can expand on if needed.

Codecov then merges these reports in the backend anyways so there is no harm if we upload them from multiple different jobs and elements in the Action as long as all of them are available after all tests are run and that seems to be the case.

Regarding docs: I moved them here in the main build action as they were failing on the main branch due to a syntax error in the MD files. To avoid this, we build the docs with every PR, but we don't deploy them (if statement in the last step). It is basically a "dry run" if the docs building and export works but the actual step to deploy them to GitHub pages only happens if the ref is the main branch and therefore only happens on merged PRs on the main branch. This ensures that we can fix docs errors or occurrences where that might fail before we even merge a PR and don't run into a regression like the one we had the last time again. Or at least let's hope so 😄

core/bluetooth/src/test/kotlin/edu/stanford/spezi/core/bluetooth/domain/BLEDeviceScannerTest.kt
```diff
@@ -86,6 +87,19 @@ class BLEDeviceScannerTest {
         assertThat(bleDeviceScanner.isScanning).isTrue()
     }
 
+    @Test
```
PSchmiedmayer, line 90: Thanks for adding this @eldcn!

PSchmiedmayer:
Thank you @eldcn for the amazing work to finish this PR up and thanks for the feedback and suggestions. I improved the state based on your feedback and provided some reasoning about some for the GitHub action setup.

Will merge this PR for now but happy to continue parts of the discussion here or in separate GitHub issues if there is still something to follow up on this PR 🚀

Issue #39, opened 2024-06-11
Handle Loading States and Errors


PSchmiedmayer:
### Problem

In the current state of the application, pressing a button (e.g. login, sign up, ...) does not indicate a loading state and does not communicate progress within the application to the user.

### Solution

We should have a reusable button + state management system (e.g. [AsyncButton](https://swiftpackageindex.com/stanfordspezi/speziviews/1.3.1/documentation/speziviews/asyncbutton) + [ViewState](https://swiftpackageindex.com/stanfordspezi/speziviews/1.3.1/documentation/speziviews/viewstate) + [ViewState Alert](https://swiftpackageindex.com/stanfordspezi/speziviews/1.3.1/documentation/speziviews/swiftui/view/viewstatealert(state:)-4wzs4) in Swift) that is easily reusable across multiple composable views and handles state management + error displays in a uniform way. Consider providing a retry mechanism (button) on error screens.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #53, opened 2024-06-30
Feature: save and load measurements

### LINKED ISSUE #52: Create mapper from FHIR Observations to Health Connect Records
### Problem

Currently, a mapper has been created in the [healthconnectonfhir](https://github.com/StanfordSpezi/SpeziKt/tree/main/modules/healthconnectonfhir) module that converts Health Connect Records into FHIR Observations. However, no mapper exists to convert FHIR Observations back into Health Connect Records.

### Solution

As discussed in our team meeting today, we will create a mapper that takes FHIR Observations and maps them to Health Connect Records by mapping the LOINC code of the Observation to the corresponding Record subclass.

### Additional context

_No response_

### Code of Conduct

Basler182:
# *save and load measurements*

## :gear: Release Notes 
image

- save measurements
- load latest measurements
- Fixes #52 

## :white_check_mark: Testing

as soon as it is confirmed that the functionality fits, further tests will be added


app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+                }
+
+                batch.commit().await()
+            }.onFailure {
+                logger.e(it) { "Error while saving observations" }
+            }
+        }
+    }
+
+    private fun getCollectionNameForObservation(observation: Observation): String {
```
PSchmiedmayer, line 72: If we pull collations, we should look into snapshot listeners to ensure that we always get the latest set of values in case something changes: https://firebase.google.com/docs/firestore/query-data/listen#kotlin+ktx

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+            observation.code.coding.any { it.code == bloodPressureCollection.code } -> bloodPressureCollection.name
+            observation.code.coding.any { it.code == heartRateCollection.code } -> heartRateCollection.name
+            else -> {
+                logger.w { "Unknown observation code" }
+                throw UnsupportedOperationException("Unknown observation code")
+            }
+        }
+    }
+
+    private suspend fun <T> getObservation(collection: ObservationCollection): Result<T?> {
```
PSchmiedmayer, line 84: Might make sense to rather rely on a snapshot listener + the overall list of collections and get the first element from there rather than multiple Firebase queries.

PSchmiedmayer (commented):
Nice job with the functionality @Basler182!

From a pure UX perspective: As discussed in our meeting today, it would be great to replace "Measured on: 30 Juni" with a concrete date and time, e.g. "June 30, 2:42 PM". In addition, the horizontal and vertical distance between the boxes should be the same.

 Nice job with the overall structure. I only had one high-level comment to use snapshot listeners for regular updates in the app; I will leave a detailed code review to @eldcn 🚀

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
Basler182: Thanks for the tip At this point it seems like it might just be overhead because the new measurements only come from the app right now. However, I will be happy to try this out with the messages feature, as the new data comes from outside the app and it will make it much easier to stay up to date at all times and when we notice something.

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/models/BluetoothUiState.kt
```diff
@@ -11,3 +13,15 @@ sealed interface BluetoothUiState {
     // Generic error
     data class Error(val message: String) : BluetoothUiState
 }
+
+data class MeasurementDialogUiState(
+    val measurement: Measurement? = null,
+    val isVisible: Boolean = false,
+    val isProcessing: Boolean = false,
```
eldcn, line 20: 💪 

Very nice. Relates to #39 and it's the correct way to handle loading states

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/component/VitalDisplay.kt
```diff
@@ -0,0 +1,169 @@
...
+        Column(
+            modifier = Modifier.padding(Spacings.medium)
+        ) {
+            Text(
+                text = vitalDisplayUiState.title,
+                style = TextStyles.titleMedium,
+                color = Colors.onBackground,
+                modifier = Modifier.testIdentifier(WeightDisplayTestIdentifier.TITLE),
+            )
+            when (vitalDisplayUiState.status) {
```
eldcn, line 55: This is again complex logic to be handled in compose. The ui states should already serve the data in such a way to be consumed directly from composables. It is the task of the mappers to construct the texts. See the existing example in `BluetoothUiStateMapper`. Instead of OperationStatus enum, I would use a statusText in the ui state instead, and `measuredOnText: String?`

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/component/VitalDisplay.kt
```diff
@@ -0,0 +1,169 @@
...
+                    Text(
+                        text = "${vitalDisplayUiState.value} ${vitalDisplayUiState.unit}",
+                        style = MaterialTheme.typography.headlineMedium,
+                        color = Colors.primary,
+                        modifier = Modifier
+                            .padding(vertical = Spacings.small)
+                            .testIdentifier(WeightDisplayTestIdentifier.VALUE),
+                    )
+                    Text(
+                        text = "Measured on: ${vitalDisplayUiState.date}",
```
eldcn, line 66: As mentioned above, this text should be constructed in the mapper. Keep also in mind string resources so that we can handle localization as well

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/mapper/MeasurementToRecordMapper.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.health.connect.client.records.Record
+import androidx.health.connect.client.records.WeightRecord
+import androidx.health.connect.client.units.Mass
+import androidx.health.connect.client.units.Pressure
+import edu.stanford.spezi.core.bluetooth.data.model.Measurement
+import java.time.LocalDateTime
+import java.time.ZonedDateTime
+import javax.inject.Inject
+
+interface MeasurementToRecordMapper {
```
eldcn, line 14: As proposed above, this functionalities can be merged in the main mapper of the view model

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/models/VitalDisplayUiState.kt
```diff
@@ -0,0 +1,24 @@
+package edu.stanford.bdh.engagehf.bluetooth.data.models
+
+import edu.stanford.bdh.engagehf.bluetooth.component.OperationStatus
+
+data class VitalDisplayUiState(
```
eldcn, line 5: This object shouldn't be needed, I would use the three properties (VitalDisplayData) directly in the main ui state of the view model instead for each type

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+                    .firstOrNull()
+                    ?.let { document ->
+                        val json = document.data
+                        val jsonString = gson.toJson(json)
+                        jsonParser.parseResource(Observation::class.java, jsonString)
+                            .toRecord() as T
+                    }
+            }.onFailure {
+                logger.e(it) { "Error while getting latest ${collection.name} observation" }
+                null
```
eldcn, line 104: this is not needed/used

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/screen/MeasurementDialog.kt
eldcn, line 1: Consider defining strings in this file and throughout this PR in general in string resources

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/screen/MeasurementDialog.kt
```diff
@@ -0,0 +1,193 @@
+@file:Suppress("MagicNumber")
```
eldcn, line 1: You can suppress the factory instead

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/screen/MeasurementDialog.kt
```diff
@@ -0,0 +1,193 @@
...
+                ) {
+                    uiState.measurement?.let {
+                        if (it is Measurement.Weight) {
+                            MeasurementRow(
+                                label = "Weight:",
+                                value = uiState.formattedWeight,
+                            )
+                        }
+                        if (it is Measurement.BloodPressure) {
+                            MeasurementRow(label = "Systolic:", value = "${it.systolic} mmHg")
```
eldcn, line 64: These texts should be served from the ui state and constructed in the mapper

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/Loinc.kt
```diff
@@ -0,0 +1,20 @@
+package edu.stanford.healthconnectonfhir
+
+object Loinc {
```
eldcn, line 3: I would be more in favor of an enum instead `enum class Loinc(code: String)`, and move `COMPONENT` in the `ObservationToRecordMapperImpl` in companion object

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/ObservationToRecordMapperImpl.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.health.connect.client.records.HeartRateRecord
+import androidx.health.connect.client.records.Record
+import androidx.health.connect.client.records.WeightRecord
+import androidx.health.connect.client.units.Mass
+import androidx.health.connect.client.units.Pressure
+import org.hl7.fhir.r4.model.Observation
+import java.time.ZoneOffset
+import javax.inject.Inject
+
+class ObservationToRecordMapperImpl @Inject constructor() : ObservationToRecordMapper {
```
eldcn, line 13: We can we mark this and `RecordToObservationMapperImpl` to internal

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/ObservationToRecordMapperImpl.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    private val codeToMappingFunction: Map<String, (Observation) -> Record> = mapOf(
+        Loinc.BloodPressure.CODE to { obs -> mapToBloodPressureRecord(obs) },
+        Loinc.Weight.CODE to { obs -> mapToWeightRecord(obs) },
+        Loinc.HeartRate.CODE to { obs -> mapToHeartRateRecord(obs) }
+    )
+
+    override fun map(observation: Observation): Record {
```
eldcn, line 21: I would introduce a generic type here:

```kotlin
   override fun <T : Record> map(observation: Observation): T {
        @Suppress("UNCHECKED_CAST")
        return when(observation.code.codingFirstRep.code) {
            Loinc.BloodPressure.CODE -> mapToBloodPressureRecord(observation)
            Loinc.Weight.CODE -> mapToWeightRecord(observation)
            Loinc.HeartRate.CODE -> mapToHeartRateRecord(observation)
            else -> error("")
        } as T
    }
```

and reflect the same generic type in the observation repository as well

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+
+    private fun Observation.toRecord(): androidx.health.connect.client.records.Record {
```
eldcn, line 109: Not needed the overhead imo, you can use `observationToRecordMapper.map(observation)` directly

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+    suspend fun getLatestBodyWeightObservation(): Result<WeightRecord?> {
+        return getObservation<WeightRecord>(bodyWeightObservation)
+    }
+
+    suspend fun getLatestHeartRateObservation(): Result<HeartRateRecord?> {
+        return getObservation<HeartRateRecord>(heartRateCollection)
+    }
+}
+
+data class ObservationCollection(val name: String, val code: String)
```
eldcn, line 126: Can be marked as private and instead of code: String, I would use the proposed enum for `Loinc` instead, type-safety ftw

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+                null
+            }
+        }
+    }
+
+    private fun Observation.toRecord(): androidx.health.connect.client.records.Record {
+        return observationToRecordMapper.map(this)
+    }
+
+    suspend fun getLatestBloodPressureObservation(): Result<BloodPressureRecord?> {
```
eldcn, line 113: The component is called observation repository, methods as well getXXXObservation, but these methods return records. I would avoid mapping to records at this component and do that only in view model via the main mapper of the view model

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+        withContext(ioDispatcher) {
+            runCatching {
+                val uid = firebaseAuth.currentUser?.uid
+                    ?: throw IllegalStateException("User not authenticated")
+                val batch = firestore.batch()
+
+                observations.forEach { observation ->
+                    val collectionName = getCollectionNameForObservation(observation)
+                    val json = jsonParser.encodeResourceToString(observation)
+                    val mapType = object : TypeToken<Map<String, Any>>() {}.type
```
eldcn, line 58: Minor: we can move this before forEach so that we only create one instance

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    suspend fun saveObservations(observations: List<Observation>) {
+        withContext(ioDispatcher) {
+            runCatching {
+                val uid = firebaseAuth.currentUser?.uid
+                    ?: throw IllegalStateException("User not authenticated")
+                val batch = firestore.batch()
+
+                observations.forEach { observation ->
+                    val collectionName = getCollectionNameForObservation(observation)
```
eldcn, line 56: Question: Should we would wrapp the content of each item here in a runCatching so that we don't miss saving the whole batch of observations in case of a single invalid one?

eldcn (commented):
The issue that we are facing now it is related to this one: https://github.com/hapifhir/hapi-fhir/issues/3268

The version lib version that we are using contains a huge file (over 80k lines) that can't be instrumented and transformed from Jacoco. I would disable androidTest coverage in app module for now and to unblock this PR and we address it separately

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -1,39 +1,73 @@
...
     private val bleService: BLEService,
     private val uiStateMapper: BluetoothUiStateMapper,
+    private val recordToObservation: RecordToObservationMapper,
+    private val observationRepository: ObservationRepository,
+    private val measurementToRecordMapper: MeasurementToRecordMapper,
 ) : ViewModel() {
     private val logger by speziLogger()
 
     private val _events = MutableSharedFlow<Event>(replay = 1, extraBufferCapacity = 1)
     private val _uiState = MutableStateFlow<BluetoothUiState>(BluetoothUiState.Idle)
```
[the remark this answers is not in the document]
Basler182: Yeah, good idea! :)

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/screen/MeasurementDialog.kt
Basler182: good idea!

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/screen/MeasurementDialog.kt
Basler182: good point! 👍

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
Basler182: well seen :)

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/repository/ObservationRepository.kt
Basler182: Theoretically, the method could also be used to store any number of independent observations. In our case, however, it is the case that related data is always saved, where it may be more intuitive not to save the others (in our case so far a maximum of 1 more) if one fails.

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/Loinc.kt
Basler182: but don't we then lose the connection of which Loinc code can have which components?

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/MeasurementModule.kt
```diff
@@ -0,0 +1,19 @@
...
+import edu.stanford.bdh.engagehf.bluetooth.data.mapper.MeasurementToRecordMapper
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+class MeasurementModule {
+
+    @Singleton
+    @Provides
+    fun provideMeasurementToRecordMapper(): MeasurementToRecordMapper =
```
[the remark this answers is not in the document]
Basler182: you are right, makes sense! 👍

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -1,39 +1,73 @@
...
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import java.time.format.DateTimeFormatter
+import java.util.Locale
 import javax.inject.Inject
 
 @HiltViewModel
 class BluetoothViewModel @Inject internal constructor(
     private val bleService: BLEService,
     private val uiStateMapper: BluetoothUiStateMapper,
```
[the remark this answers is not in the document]
Basler182: here i'm not quite sure, because not all mappers have anything to do with the ui state. in the ui state we show the "measurement" and some mappers we only use to map the measurement into a record or observation so that we can save it, but don't use it for visualisation. or would you still include it in the ui state mapper, even if it's not directly related to the ui state, @eldcn ?

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -51,21 +85,261 @@ class BluetoothViewModel @Inject internal constructor(
...
+            is Action.ConfirmMeasurement -> {
+                _dialogUiState.update {
+                    it.copy(isProcessing = true)
+                }
+                viewModelScope.launch {
+                    if (action.measurement is Measurement.Weight) {
+                        val records = measurementToRecordMapper.map(action.measurement)
+                        records.forEach { record ->
+                            recordToObservation.map(record).let {
+                                observationRepository.saveObservations(it)
```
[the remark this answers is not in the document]
Basler182: after talking to @PSchmiedmayer , i have slightly changed the logic. we no longer just query the data, but have a listener that notifies us of changes. i have made the optimisations with the mapper. i like it, then the viemodel will be a little bit slimmer again :)

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/ObservationToRecordMapperImpl.kt
Basler182: Good idea. I like this approach. 💯

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/ObservationToRecordMapperImpl.kt
Basler182: I think this is not possible cause in `HealthConnectOnFhirModule`  it would say that public function exposes its internal parameter type.

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/component/VitalDisplay.kt
Basler182: but without the enum it feels even more complex. then we have to create logic when e.g. the CircularProgressBar is displayed if e.g. Status.isEmpty and text.isEmpty, etc?

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/mapper/MeasurementToRecordMapper.kt
Basler182: logic moved into repository

eldcn:
@PSchmiedmayer Unfortunately, the fixed that I added still didn't work as there was a crash at runtime for missing classpath. We decided to disable coverage for android tests for now, and will think for an alternative solution in a follow up. 

Also, there will be a follow up to address some smaller issues to not block this PR. 

Since the code coverage decreased, the code-cov job fails, and we might need your support to merge directly. This iteration looks very good so far 👍

eldcn (approved):

Basler182:
@PSchmiedmayer and we can make the changes due to the firestore changes in a small follow up pr, since not much changes, it should be done quickly

PSchmiedmayer:
@Basler182 @eldcn Sounds good, thank you for all the work here! 

Please note any remaining issues in GitHub issues and let me know when I should merge the PR.

Also fine to address the changes in paths in a separate, smaller PR as long is both systems don't stay out of sync for too long.

Basler182:
> @Basler182 @eldcn Sounds good, thank you for all the work here!
> 
> Please note any remaining issues in GitHub issues and let me know when I should merge the PR.
> 
> Also fine to address the changes in paths in a separate, smaller PR as long is both systems don't stay out of sync for too long.

we are ready to go. you are welcome to merge it, @PSchmiedmayer  🚀

PSchmiedmayer:
Sounds great; thank you for the work @Basler182 @eldcn!

PR #69, opened 2024-07-22
adjust firestore security rules from users to patients

Basler182:
# *adjust firestore security rules from users to patients*

[Adapt FirebaseStorage Rules for users->patients renaming PR](https://github.com/StanfordBDHG/ENGAGE-HF-Firebase/pull/33)
[Add security](https://github.com/StanfordBDHG/ENGAGE-HF-Firebase/pull/24)

### DIFF
--- a/modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
+++ b/modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
@@ -78,7 +78,7 @@ internal class AuthenticationManager @Inject constructor(
                     "gender" to selectedGender,
                     "dateOfBirth" to dateOfBirth
                 )
-                firestore.collection("users").document(user.uid).set(userMap).await()
+                firestore.collection("patients").document(user.uid).set(userMap).await()
             }
         }.onFailure { e ->
             logger.e { "Error saving user data: ${e.message}" }
--- a/modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
+++ b/modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
@@ -84,7 +84,7 @@ internal class AuthenticationManager @Inject constructor(
                     "GenderIdentityKey" to GenderIdentity.fromDisplayName(selectedGender).databaseName,
                     "DateOfBirthKey" to birthDayTimestamp
                 )
-                firestore.collection("patients").document(user.uid).set(userMap).await()
+                firestore.collection("users").document(user.uid).set(userMap).await()
             }
         }.onFailure { e ->
             logger.e { "Error saving user data: ${e.message}" }
--- a/modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
+++ b/modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
@@ -2,14 +2,17 @@
 
 package edu.stanford.spezi.module.account.manager
 
+import com.google.firebase.Timestamp
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.auth.FirebaseUser
 import com.google.firebase.auth.GoogleAuthProvider
 import com.google.firebase.auth.UserProfileChangeRequest
 import com.google.firebase.firestore.FirebaseFirestore
 import edu.stanford.spezi.core.logging.speziLogger
+import edu.stanford.spezi.module.account.register.GenderIdentity
 import kotlinx.coroutines.tasks.await
 import java.time.LocalDate
+import java.time.ZoneId
 import javax.inject.Inject
 
 internal class AuthenticationManager @Inject constructor(
@@ -33,11 +36,13 @@ internal class AuthenticationManager @Inject constructor(
         return firebaseAuth.currentUser
     }
 
-    private suspend fun setDisplayName(firstName: String, lastName: String) {
+    private suspend fun setUserProfile(firstName: String, lastName: String, email: String) {
         val user = firebaseAuth.currentUser
         val profileUpdates = UserProfileChangeRequest.Builder()
             .setDisplayName("$firstName $lastName")
             .build()
+
+        user?.updateEmail(email)?.await()
 
         user?.updateProfile(profileUpdates)?.await()
     }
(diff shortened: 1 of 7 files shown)


modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
```diff
@@ -32,19 +32,20 @@ internal class UserSessionManagerImpl @Inject constructor(
...
-                .putStream(inputStream)
-                .await().task.isSuccessful
+    override suspend fun uploadConsentPdf(pdfBytes: ByteArray): Result<Unit> =
+        withContext(ioDispatcher) {
+            runCatching {
+                val currentUser = firebaseAuth.currentUser ?: error("User not available")
+                val inputStream = ByteArrayInputStream(pdfBytes)
+                logger.i { "Uploading file to Firebase Storage" }
+                val uploaded = firebaseStorage
+                    .getReference("patients/${currentUser.uid}/consent/signature.pdf")
```
PSchmiedmayer, line 42: On iOS we currently store it under `consent/\(dateString).pdf` ("yyyy-MM-dd_HHmmss" as the date string format): https://github.com/StanfordBDHG/ENGAGE-HF-iOS/blob/main/ENGAGEHF/ENGAGEHFStandard.swift#L193. 

We should use the same logic here as well for now. We might want to adjust this in the future.

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
```diff
@@ -72,7 +73,7 @@ internal class UserSessionManagerImpl @Inject constructor(
     private suspend fun hasConsented(): Boolean = withContext(ioDispatcher) {
         runCatching {
             val uid = getUserUid() ?: error("No uid available")
-            val reference = firebaseStorage.getReference("users/$uid/signature.pdf")
+            val reference = firebaseStorage.getReference("patients/$uid/consent/signature.pdf")
```
PSchmiedmayer, line 76: See comment above.

modules/account/src/test/java/edu/stanford/spezi/module/account/manager/UserSessionManagerTest.kt
```diff
@@ -225,7 +225,7 @@ class UserSessionManagerTest {
 
     private fun setupConsented() {
         val uid = "uid"
-        val location = "users/$uid/signature.pdf"
+        val location = "patients/$uid/consent/signature.pdf"
```
PSchmiedmayer, line 228: See comment above.

modules/account/src/test/java/edu/stanford/spezi/module/account/manager/UserSessionManagerTest.kt
```diff
@@ -244,7 +244,7 @@ class UserSessionManagerTest {
 
     private fun setupPDFUpload(successful: Boolean) {
         val uid = "uid"
-        val location = "users/$uid/signature.pdf"
+        val location = "patients/$uid/consent/signature.pdf"
```
PSchmiedmayer, line 247: See comment above.

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
```diff
@@ -78,7 +78,7 @@ internal class AuthenticationManager @Inject constructor(
                     "gender" to selectedGender,
                     "dateOfBirth" to dateOfBirth
                 )
-                firestore.collection("users").document(user.uid).set(userMap).await()
+                firestore.collection("patients").document(user.uid).set(userMap).await()
```
PSchmiedmayer, line 81: @pauljohanneskraft, @Supereg: We will need to eventually clearly define the keys for user information stored in Spezi Account / SpeziFirebase in a documentation (e.g. inline docs or README) and we need to sync that up between iOS & Android.

@Basler182 Currently email and names are not stored in firebase as they are stored in firebase Authentication, we should do that the same way here: https://firebase.google.com/docs/auth/android/manage-users#get_a_users_profile

In addition on iOS we store gender & dob using the following keys:
- `DateOfBirthKey`. Date of both needs to be a Firebase datetime.
- `GenderIdentityKey`. Is stored as a string.
We should document this in the Firebase repo @pauljohanneskraft.

I think both could be improved in the future, that might be an open ticket and something to consider in the future @Supereg & @nriedman.

modules/measurements/src/main/kotlin/edu/stanford/spezi/modules/measurements/MeasurementsRepository.kt
PSchmiedmayer, line 1: Thank you for moving this over to the patients collection!

PSchmiedmayer (approved):
Thank you for the improvements @Basler182!

I had a few comments; once they are addressed I would be happy to see this merged 🚀

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
Basler182: @PSchmiedmayer  I think it would make sense to omit the ‘Key’ at the end of the field names and start with a lower case letter as we do in the other places as far as I know. I can update the email at this point as a precaution. We save the names in the user profile at this point anyway. I'll remove the double save in the Firestore. 

For the gender, it would be good if we had a list with the possible choices so that it is identical on iOS and Android. @pauljohanneskraft

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
Basler182: We are currently checking whether the file has already been signed. If we make the change to the date naming, we will no longer know the name of the file at a later date and could only check whether there is any file in the folder. The date feels redundant because we save the creation and update date of the file anyway.  This could also lead to problems if we later have several consent files. However, if this is wanted, then I can adjust it @PSchmiedmayer  :)

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
PSchmiedmayer: Sure, here is the list of supported types: https://github.com/StanfordSpezi/SpeziAccount/blob/e3344aaf5119e124d96aed39e04e4ce3f035ae41/Sources/SpeziAccount/Model/GenderIdentity.swift#L13

And I would leave the keys as they are for now; we should make this configurable & modifiable in a nice way in Spezi Account in the future @Supereg.

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
PSchmiedmayer: To recap the discussion in the meeting today: We will use a `/patients/{PATIENT_ID}/consent.pdf` path without any version numbers nor a lot of nesting for now. This needs to be aligned with the iOS app @nriedman.

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
PSchmiedmayer: Yes, that's correct. 00:00 at the time of the day is correct.

PR #70, opened 2024-07-28
Authentication components

### LINKED ISSUE #61: Education Page Dark Mode Improvements
### Description

The education page is duplicated on the medication page and is not properly rendering the background of the cards in dark mode on an android device.

<p float="left">
  
  
</p>

### Reproduction

Open the application in dark mode on a recent android device.

### Expected behavior

It would be great if the cards would have a slight contrast to the background (e.g. dark gray) to separate them from the background while being distinct. Android might have a secondary background color style. It might make sense to define this across the ENGAGE-HF app and maybe even within the Spezi Core UI module.

### Additional context

_No response_

### Code of Conduct

eldcn:
# *Authentication components*

- Started as a Fix for #61, but when debugging I noticed that even on login failure, we ignore the error and navigate to app screen.
- Issue was because in auth manager we were returning `Result<Boolean>`, and were not checking the value of the result, but only it's `.isSuccess`, while boolean value was also containing the info whether a certain operation was successful or not. 
  - Unified this behaviour by using `Result<Unit>` in auth related components
  - Removed `CredentialLoginManagerAuth` and `CredentialRegisterManagerAuth` as they were solely wrapping `AuthenticationManager`, and used instead `AuthenticationManager` directly in the view models
- Added missing unit tests for components visited during the changes
- Finally, regarding the actual fix #61, it looked like @Basler182 has addressed the issue in a former PR. I noticed however the elevation color in dark mode, was in fact using light mode. Reason was because in same cases we were passing `.lighten(isSystemInDarkTheme)` and in other cases the method was using the default value which was `false`. Fixed by marking `Color.lighten()` extension method with `@Composable` and access dark mode flag in the method itself:

**Before:**

image

**After:**
image

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing


modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
```diff
@@ -49,69 +57,105 @@ internal class AuthenticationManager @Inject constructor(
...
+            val request = GetCredentialRequest.Builder()
+                .addCredentialOption(googleIdOption)
+                .build()
+
+            val credential = credentialManager.getCredential(
+                request = request,
+                context = context
+            ).credential
+            if (credential is CustomCredential && credential.type == GoogleIdTokenCredential.TYPE_GOOGLE_ID_TOKEN_CREDENTIAL) {
+                GoogleIdTokenCredential.createFrom(credential.data)
```
eldcn, line 149: This method is not tested because of this usage, mockk could not stub the static behaviour on this component

PSchmiedmayer (commented):
Thank you @eldcn; would be great if you can review this in detail @Basler182. 

Happy to see this merged, thanks for all the work @eldcn!

core/design/src/main/kotlin/edu/stanford/spezi/core/design/theme/Colors.kt
```diff
@@ -87,7 +88,9 @@ internal val RectangleBlue = Color(0xFFEBF2FC)
 private const val DARK_FACTOR = 0.1f
 private const val LIGHT_FACTOR = 0.9f
 
-fun Color.lighten(isDarkTheme: Boolean = false): Color {
+@Composable
+fun Color.lighten(): Color {
+    val isDarkTheme = isSystemInDarkTheme()
```
Basler182, line 93: the idea to convert the lighten into a composeable and call isSysteminDarkTheme directly by default is great 🚀

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
```diff
@@ -49,69 +57,105 @@ internal class AuthenticationManager @Inject constructor(
         lastName: String,
         selectedGender: String,
         dateOfBirth: LocalDate,
-    ): Result<Boolean> {
+    ): Result<Unit> {
```
Basler182, line 60: great that you noticed this!

Basler182 (approved):
Great changes. Approved! 💯

PR #91, opened 2024-08-19
Feature/questionnaire

Basler182:
# *Feature/questionnaire*

questionnaire wasnt implemented

## :gear: Release Notes 
- Questionnaire Action
- show Questionnaire
- Load Questionnaire 
- Save QuestionnaireResponse

Known problems: 

- The keyboard is partially displayed, which theoretically allows you to insert text into the fragments. It is prevented that you can send wrong repsonses, but it would be better if they were prevented from the beginning. 
- It could be possible to switch to RadioButton instead of DropdownTextFields to prevent the behaviour
- Perhaps we could also integrate a keyboard listener and automatically remove the keyboard when it opens to prevent someone from entering free text. This is possible because we do not have any free text questions in our questionnaires
- Its not possible to show the page in a bottom sheet like the others, cause of this bug: https://issuetracker.google.com/issues/277651136?pli=1

### DIFF
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
@@ -74,6 +74,7 @@ class MainActivity : FragmentActivity() {
                             navHostController = navHostController,
                             startDestination = uiState.startDestination
                         )
+                        setTheme(R.style.Theme_Spezi_Content)
                     }
                 }
             }
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
@@ -136,7 +136,11 @@ class MainActivity : FragmentActivity() {
             AppScreen()
         }
 
-        composable<Routes.QuestionnaireScreen> {
+        composable<Routes.QuestionnaireScreen>(
+            typeMap = mapOf(
+                typeOf<String>() to serializableType<String>()
+            )
+        ) {
             QuestionnaireScreen()
         }
 
@@ -173,7 +177,9 @@ class MainActivity : FragmentActivity() {
                             )
                         )
 
-                        is AppNavigationEvent.QuestionnaireScreen -> navHostController.navigate(Routes.QuestionnaireScreen)
+                        is AppNavigationEvent.QuestionnaireScreen -> navHostController.navigate(
+                            Routes.QuestionnaireScreen(event.questionnaireId)
+                        )
 
                         is AccountNavigationEvent.LoginScreen -> navHostController.navigate(
                             Routes.LoginScreen(
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
@@ -1,7 +1,6 @@
 package edu.stanford.bdh.engagehf
 
 import android.os.Bundle
-import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
 import androidx.compose.foundation.layout.Box
@@ -14,6 +13,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
+import androidx.fragment.app.FragmentActivity
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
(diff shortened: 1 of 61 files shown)


app/src/main/kotlin/edu/stanford/bdh/engagehf/questionnaire/QuestionnaireRepository.kt
```diff
@@ -0,0 +1,75 @@
...
+class QuestionnaireRepository @Inject constructor(
+    @Dispatching.IO private val ioDispatcher: CoroutineDispatcher,
+    private val questionnaireDocumentMapper: QuestionnaireDocumentMapper,
+    private val observationCollectionProvider: ObservationCollectionProvider,
+    private val firestore: FirebaseFirestore,
+) {
+    private val logger by speziLogger()
+
+    suspend fun observe(id: String): Flow<Result<Questionnaire>> = callbackFlow {
+        val listenerRegistration: ListenerRegistration? = null
```
eldcn, line 29: should be a `var` and we should assign it below so that we can clear

app/src/main/kotlin/edu/stanford/bdh/engagehf/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+    private val navigator: Navigator,
+    savedStateHandle: SavedStateHandle,
+    private val notifier: MessageNotifier,
+    private val jsonParser: IParser,
+) : ViewModel() {
+    private val logger by speziLogger()
+
+    private val _uiState = MutableStateFlow<State>(
+        State.Loading(
+            savedStateHandle.decode(QUESTIONNAIRE_SAVE_STATE_PARAM)
```
eldcn, line 36: We can keep the the id as a private property or extend the `loadQuestionnaire(id = savedStateHandle.decode(KEY)` in init block

app/src/main/kotlin/edu/stanford/bdh/engagehf/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+    init {
+        loadQuestionnaire()
+    }
+
+    private fun loadQuestionnaire() {
+        viewModelScope.launch {
+            val questionnaireId = (_uiState.value as? State.Loading)?.questionnaireId
+            questionnaireId?.let {
+                val questionnaire =
+                    questionnaireRepository.observe(it).first()
```
eldcn, line 50: If we are just interested in the first emission, is it really required to have observe on repository? Isn't it sufficient to simply get once the questionnaire by id and return a kotlin.Result?

It would be great to also handle error case as well. currently the app would crash in case something goes wrong with internet connection, no questionnaire from firebase or if encoding fails

app/src/main/kotlin/edu/stanford/bdh/engagehf/questionnaire/QuestionnaireViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+        when (action) {
+            is Action.SaveQuestionnaireResponse -> {
+                val response = action.response
+                response.setAuthored(Date())
+                logger.i { "Save questionnaire response: $response" }
+                viewModelScope.launch {
+                    val questionnaire = _uiState.value as? State.QuestionnaireLoaded
+                        ?: error("Invalid state")
+                    _uiState.update {
+                        State.Loading(null)
```
eldcn, line 76: As proposed above I would extract the id from the loading state

app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
```diff
@@ -48,7 +49,7 @@ import javax.inject.Inject
 import kotlin.reflect.typeOf
 
 @AndroidEntryPoint
-class MainActivity : ComponentActivity() {
+class MainActivity : FragmentActivity() {
```
eldcn, line 52: Can you please also update the `ComposeContentActivity` to extend `FragmentActivity` as well? We use it for testing and it would be great to have the same behaviour

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -162,7 +163,12 @@ class BluetoothViewModel @Inject internal constructor(
                             is MessagesAction.MedicationsAction -> { /* TODO */
                             }
 
-                            is MessagesAction.QuestionnaireAction -> { /* TODO */
+                            is MessagesAction.QuestionnaireAction -> {
```
eldcn, line 166: Can we just pass the id in the action?

app/src/main/kotlin/edu/stanford/bdh/engagehf/questionnaire/QuestionnaireViewModel.kt
Basler182: Yea this makes a lot of sense. I changed it.

eldcn (approved):

PR #96, opened 2024-09-08
Feature/notification

Basler182:
# *Feature notification*

No notifications are used right now.

## :gear: Release Notes 

The notification module provides a way to show notifications to the user and handle FCM as well as
the notification settings.

## :books: Documentation
- The `NotificationSetting` is responsible for handling the notification settings. It provides a way to enable or disable notifications.
- The `DeviceRegistrationService` is responsible for registering the device to receive FCM notifications and the `FCMNotificationService` is responsible for handling the FCM notifications.
- `NotificationNotifier` is an interface that defines the contract for a notifier responsible for showing notifications.

## :white_check_mark: Testing
The tests are added after the first review.


core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettingsRepository.kt
```diff
@@ -0,0 +1,112 @@
...
+                runCatching {
+                    val userId = userSessionManager.getUserUid() ?: error("User not authenticated")
+                    listenerRegistration = firestore.collection(NOTIFICATION_SETTINGS_PATH)
+                        .document(userId)
+                        .addSnapshotListener { snapshot, exception ->
+                            if (exception != null) {
+                                logger.e(exception) { "Error observing notification settings" }
+                                trySend(Result.failure(exception))
+                            } else {
+                                val notificationSettings =
```
eldcn, line 39: I would propose an alternative data structure that is a bit more scalable imo:

```kotlin
enum class NotificationType(val key: String) {
    APPOINTMENT_REMINDERS(key = "receivesAppointmentReminders"),
    MEDICATION_UPDATES(key = "receivesMedicationUpdates"),
    QUESTIONNAIRE_REMINDERS(key = "receivesQuestionnaireReminders"),
    RECOMMENDATION_UPDATES(key = "receivesRecommendationUpdates"),
    VITALS_REMINDERS(key = "receivesVitalsReminders"),
    WEIGHT_ALERTS(key = "receivesWeightAlerts"),
}

data class NotificationSettings(
    private val settings: Map<NotificationType, Boolean>
) : Map<NotificationType, Boolean> by settings {

    override fun get(key: NotificationType): Boolean {
        return settings[key] ?: false
    }
    
    fun update(type: NotificationType, value: Boolean): NotificationSettings {
        val newSettings = settings.toMutableMap()
        newSettings[type] = value
        return NotificationSettings(settings = newSettings.toMap())
    }
}
```

This snippet could be replaced to:

```kotlin
val notificationSettings = NotificationSettings(
    settings = NotificationType.entries.associateWith { type ->
        snapshot?.getBoolean(type.key) ?: false
    }
)
trySend(Result.success(notificationSettings))
```

and in save settings:

```kotlin
suspend fun saveNotificationSettings(
    notificationSettings: NotificationSettings,
): Result<Unit> {
    return withContext(ioDispatcher) {
        runCatching {
            val userId = userSessionManager.getUserUid() ?: error("User not authenticated")
            firestore.collection(NOTIFICATION_SETTINGS_PATH)
                .document(userId)
                .update(notificationSettings.mapKeys { it.key.key })
                .await().let { }
        }
    }
}
```

This can replace the need for the constants, for the mapper for switch type enum.

Feel free to extend the `NotificationType` enum with an additional property for the section (Reminders, updates or Trends), and adjust the ui model to group the settings by sections and use a lazy column in compose. With these adjustments, I think adding or deleting a new setting requires only changing the `NotificationType` and the rest will behave the same

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/FCMNotificationService.kt
```diff
@@ -0,0 +1,65 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+/*
+ * Service to handle incoming notifications from Firebase Cloud Messaging (FCM)
+ * and register the device with the server when a new token is generated.
+ */
+@AndroidEntryPoint
+class FCMNotificationService @Inject internal constructor(
```
eldcn, line 23: This is an android component (android service) and similar to activities we can't initialize or inject directly, but the initialization happens from OS e.g. via `context.startService`, hence `@Inject internal constructor` is not needed

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -54,6 +56,7 @@ class BluetoothViewModel @Inject internal constructor(
         observeBleService()
         observeRecords()
         observeMessages()
+        viewModelScope.launch { deviceRegistrationService.registerDevice() }
```
eldcn, line 59: I would not mark the `registerDevice` as suspend and launch  `ioScope` internally instead. The reason is that we want that operation to finish and not bound it to the lifecycle of this view model.

QQ: Is this manual call necessary? Can't we just rely on `onNewToken` in the service? From the docs: `Called when a new token for the default Firebase project is generated.
This is invoked after app install when a token is first generated, and again if the token changes.` The `DeviceRegistrationService` can then also be marked internal. We can then also update it to `registerDevice(token: String)` and call it with the new token, `MessageTokenService` can also be removed then as not needed

app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
```diff
@@ -53,6 +65,10 @@ class MainActivity : FragmentActivity() {
 
     private val viewModel by viewModels<MainActivityViewModel>()
 
+    private val bluetoothViewModel by viewModels<BluetoothViewModel>()
```
eldcn, line 68: Bluetooth (Home) view model does a lot more and it's associated with the home screen. I would create a new component, `MessageHandler` and extract the logic that it is currently happening in the bluetooth view model into that component. That component can then be injected in the `MainActivityViewModel` and here we can invoke a new action that will internally invoke the handler

core/notification/src/main/java/edu/stanford/spezi/core/notification/di/NotificationModule.kt
```diff
@@ -0,0 +1,37 @@
...
+import edu.stanford.spezi.core.notification.notifier.SystemTrayNotificationNotifier
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+internal class NotificationModule {
+    @Provides
+    @Singleton
+    internal fun provideFirebaseMessaging(): FirebaseMessaging =
+        FirebaseMessaging.getInstance().apply {
```
eldcn, line 21: can we remove the empty apply block or are you planning any further configuration?

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    private suspend fun getNotificationToken(): Result<String?> {
+        return runCatching {
+            messageTokenService.getNotificationToken()
+        }.onFailure { e ->
+            logger.e(e) { "Error getting notification token" }
+        }
+    }
+
+    private fun getAppVersion(): String {
```
eldcn, line 59: Would update to one single `private fun getPackageInfo() = context.packageManager.getPackageInfo(context.packageName, 0)` and use `packageInfo.versionName` and `packageInfo.versionCode` above

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/MessagesAction.kt
```diff
@@ -6,6 +6,7 @@ sealed class MessagesAction {
     data object MeasurementsAction : MessagesAction()
     data class QuestionnaireAction(val questionnaireId: String) : MessagesAction()
     data object HealthSummaryAction : MessagesAction()
+    data object NoAction : MessagesAction()
```
eldcn, line 9: Minor: I am more in favor of nullable instead of introducing a new type that will always stay in memory for nothing 😄

core/notification/README.MD
```diff
@@ -0,0 +1,28 @@
...
+
+## FCM
+
+The `DeviceRegistrationService` is responsible for registering the device to receive FCM
+notifications and the `FCMNotificationService` is responsible for handling the FCM notifications.
+
+## Setting
+
+The `NotificationSetting` is responsible for handling the notification settings. It provides a way
+to enable or disable notifications.
```
eldcn, line 28: Please review again all components from this module that should be public. If I am not mistaken, only the compose screen, proposed DI qualifier for component name, `FirebaseMessage`, navigation event, and route should be public.

app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
```diff
@@ -80,6 +96,25 @@ class MainActivity : FragmentActivity() {
                 }
             }
         }
+        logger.i { "Intent Data from Main Activity ${intent.extras}" }
```
eldcn, line 99: This will work in case the app did not launch yet. In case the main activity has started and the app is in background we will create a new instance of it. I propose the following. Update the `android:launchMode="singleInstance"` for the `MainActivity` in `AndroidManifest.xml`, and override here `onNewIntent` method, from the docs of `onNewIntent`

> Only allow one instance of this activity to ever be running. This activity gets a unique task with only itself running in it; if it is ever launched again with the same Intent, then that task will be brought forward and its {@link android. app. Activity#onNewIntent Activity. onNewIntent()} method called. If this activity tries to start a new activity, that new activity will be launched in a separate task. See the Tasks and Back Stack document for more details about tasks.

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/FCMNotificationService.kt
Basler182: now i understand why i couldn't inject anything in the constructor, thanks for the explanation 🚀

core/notification/src/main/java/edu/stanford/spezi/core/notification/di/NotificationModule.kt
Basler182: i thought i could add a debug emulator setting similar to the other firebase components but the component doesn't seem to support it and then i must have forgotten the apply block. i removed it

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
Basler182: I called it on every startup because the RegisterDevice method also updates other things besides the token, such as the language and time zone. But I guess these are things that change so rarely that it is negligible. i have adapted it according to the suggestion

core/notification/src/main/java/edu/stanford/spezi/core/notification/notifier/SystemTrayNotificationNotifier.kt
```diff
@@ -0,0 +1,111 @@
...
+                    notification.hashCode(),
+                    notification,
+                )
+            }
+        }
+
+    companion object {
+        const val SPEZI_MESSAGE_NOTIFICATION_CHANNEL_ID = "SPEZI_MESSAGE_NOTIFICATION_CHANNEL"
+        const val SPEZI_MESSAGE_NOTIFICATION_REQUEST_CODE = 0
+        const val TARGET_ACTIVITY_NAME = "edu.stanford.spezi.core.MainActivity"
```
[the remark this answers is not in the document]
Basler182: injecting the ComponentName in that way is much better than the static non type-safe one i did. i like the one with the annotation. thanks for the tip. i changed it 🚀

app/src/main/AndroidManifest.xml
```diff
@@ -9,15 +8,16 @@
         android:fullBackupContent="@xml/backup_rules"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
+        android:launchMode="singleInstance"
```
eldcn, line 11: should be done below for the main activity and not for the application 🙌

Basler182: oops well seen

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
eldcn: What I meant here was to invoke `getPackageInfo()` just once:

```kotlin
val packageInfo = getPackageInfo()
...
val deviceInfo = DeviceInfo(
            notificationToken = token,
            platform = platform,
            osVersion = osVersion,
            appVersion = packageInfo.versionName,
            appBuild = packageInfo.versionCode.toString(),
            language = language,
            timeZone = timeZone
        )
```

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
```diff
@@ -0,0 +1,108 @@
...
+                KEY_OS_VERSION to osVersion,
+                KEY_APP_VERSION to appVersion,
+                KEY_APP_BUILD to appBuild,
+                KEY_LANGUAGE to language,
+                KEY_TIME_ZONE to timeZone
+            )
+        }
+
+        runCatching {
+            CoroutineScope(ioDispatcher).launch {
```
eldcn, line 75: Can we inject `@Dispatching.IO private val coroutineScope: CoroutineScope` directly instead of dispatcher?

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
```diff
@@ -0,0 +1,108 @@
...
+        private const val KEY_PLATFORM = "platform"
+        private const val KEY_OS_VERSION = "osVersion"
+        private const val KEY_APP_VERSION = "appVersion"
+        private const val KEY_APP_BUILD = "appBuild"
+        private const val KEY_LANGUAGE = "language"
+        private const val KEY_TIME_ZONE = "timeZone"
+    }
+}
+
+data class DeviceInfo(
```
eldcn, line 100: Can be private, or we can can also build the map directly and remove this class?

core/notification/src/main/java/edu/stanford/spezi/core/notification/fcm/FCMNotificationService.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Called when a message is received from Firebase Cloud Messaging.
+     * Example remoteMessage.data:
+     * {messageId=XYZID, action=observations, type=Vitals, isDismissible=true}
+     */
+    override fun onMessageReceived(remoteMessage: RemoteMessage) {
+        logger.i { "data: ${remoteMessage.data}" }
+        remoteMessage.notification?.let { notification ->
+            notificationNotifier.sendNotification(
+                firebaseMessage = FirebaseMessage(
+                    title = notification.title ?: "New ENGAGE-HF Message",
```
eldcn, line 36: Let's remove `ENGAGE-HF` from the default string?

core/notification/src/main/java/edu/stanford/spezi/core/notification/notifier/NotificationNotifier.kt
```diff
@@ -0,0 +1,93 @@
...
+import androidx.core.content.PermissionChecker
+import dagger.hilt.android.qualifiers.ApplicationContext
+import edu.stanford.spezi.core.logging.speziLogger
+import edu.stanford.spezi.core.notification.R
+import javax.inject.Inject
+
+/**
+ * A class that sends notifications using the system tray.
+ */
+class NotificationNotifier @Inject constructor(
```
eldcn, line 22: Can we mark this as internal and move the message key in a companion object of `FirebaseMessage`?

core/notification/src/main/java/edu/stanford/spezi/core/notification/notifier/NotificationNotifier.kt
```diff
@@ -0,0 +1,93 @@
...
+            )
+        }
+    }
+
+    private fun createMessageNotification(
+        firebaseMessage: FirebaseMessage,
+    ): Notification {
+        return NotificationCompat.Builder(context, SPEZI_MESSAGE_NOTIFICATION_CHANNEL_ID)
+            .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+            .setSmallIcon(edu.stanford.spezi.core.design.R.drawable.ic_info).setContentTitle(
```
eldcn, line 52: Minor: Can we move each `.set` method in a new line?

core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettingScreen.kt
```diff
@@ -0,0 +1,200 @@
...
+}
+
+@Composable
+private fun NotificationOptions(
+    notificationSettings: NotificationSettings,
+    onAction: (NotificationSettingViewModel.Action) -> Unit,
+    pendingActions: PendingActions<NotificationType>,
+) {
+    LazyColumn(modifier = Modifier.padding(vertical = Spacings.medium)) {
+        notificationSettings.groupBySection().forEach { (section, settings) ->
```
eldcn, line 106: Can we do this above using remember? I would move the extension function here in this file

```kotlin
@Composable
private fun NotificationSettings.groupBySection(): Map<NotificationType.Section, List<Pair<NotificationType, Boolean>>> {
    return remember(key1 = this) {
        entries.groupBy(
            keySelector = { it.key.section },
            valueTransform = { it.toPair() }
        )
    }
}
```

core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettingViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+                    _uiState.update {
+                        UiState.NotificationSettingsLoaded(
+                            currentSettings.copy(
+                                pendingActions = currentSettings.pendingActions.plus(
+                                    action.notificationType
+                                )
+                            )
+                        )
+                    }
+                    viewModelScope.launch {
```
eldcn, line 67: Can we remove the action once?

```kotlin
viewModelScope.launch {
    repository.saveNotificationSettings(newSettings)
        .onFailure {
            messageNotifier.notify("Failed to save notification settings")
        }
    _uiState.update {
        UiState.NotificationSettingsLoaded(
            currentSettings.copy(
                pendingActions = currentSettings.pendingActions.minus(
                    action.notificationType
                )
            )
        )
    }
}
```

core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettingViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    sealed interface UiState {
+        data object Loading : UiState
+        data class Error(val message: String) :
+            UiState
+
+        data class NotificationSettingsLoaded(
+            val notificationSettings: NotificationSettings,
+        ) :
+            UiState
```
eldcn, line 114: Minor: Would be great to declare the conformance in the same line with `:` here and above, basically just move `UiState` above

core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettings.kt
```diff
@@ -0,0 +1,43 @@
+package edu.stanford.spezi.core.notification.setting
+
+import edu.stanford.spezi.core.design.action.PendingActions
+
+internal enum class NotificationType(
```
eldcn, line 5: 👏

core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettings.kt
```diff
@@ -0,0 +1,43 @@
...
+    ;
+
+    enum class Section {
+        REMINDERS, UPDATES, TRENDS
+    }
+}
+
+internal data class NotificationSettings(
+    private val settings: Map<NotificationType, Boolean>,
+    val pendingActions: PendingActions<NotificationType> = PendingActions(),
```
eldcn, line 24: I would move this in the `NotificationSettingsLoaded` type instead and use `SwitchChanged` as the action type

core/notification/src/main/java/edu/stanford/spezi/core/notification/setting/NotificationSettingsStateMapper.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Maps [NotificationSettingViewModel.Action] to [NotificationSettingViewModel.UiState].
+ */
+internal class NotificationSettingsStateMapper @Inject constructor() {
+    fun mapSwitchChanged(
+        action: NotificationSettingViewModel.Action.SwitchChanged,
+        currentSettings: NotificationSettings,
+    ): NotificationSettings {
+        val updatedSettings = currentSettings.update(action.notificationType, action.isChecked)
+        return updatedSettings.copy(pendingActions = currentSettings.pendingActions + action.notificationType)
```
eldcn, line 14: As proposed in the other comment, I would move the pending actions as a property of the ui state. Also here this operation is not being used as the mapped settings are just used to be saved in firebase. you are again updating the pending actions in the view model on current settings

eldcn (approved):
Great job 🚀 I left some additional minor comments :)

PR #104, opened 2024-09-19
SpeziContact

pauljohanneskraft:
# SpeziContact

The existing Contact module doesn't follow the iOS interface and we may want to align it more.

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing


pauljohanneskraft:
We are currently still missing ContactsList

Basler182:
These are just a few thought experiments intended to drive the discussion forward and introduce new perspectives. I think it would also be possible, for example, to use the following:

Create a interface:
```kotlin
interface ContactOptionType {
    val title: String
    fun handleAction(context: Context)
}
```

Provide custom implementations:
```kotlin
class WebsiteContactOptionType(override val title: String) : ContactOptionType {
    override fun handleAction(context: Context) {
        // create and start intent
    }
}
```
Create a Factory Method in a Companion Object for Default ones. 

Handle the action in ViewModel like this:
```kotlin
@HiltViewModel
class ContactViewModel @Inject internal constructor(
    @ApplicationContext private val context: Context,
) : ViewModel() {
    fun onAction(action: Action) {
        is OnAction.CardClicked -> {
             action.option.optionType.handleAction(context)
         }
}
```
And the Context could be DI in the ViewModel.

pauljohanneskraft:
Thank you for your input @Basler182!

I agree that this is a great way to avoid the lambda solution. What I'm wondering about is whether it is possible to make it easier to identify which options are available to a potential developer. With the `ContactOption.call(...)` function, for example, we made it possible to simply check all the methods of the companion object to find available options. In Swift, this would be possible, so no issue there, I was just wondering about Kotlin/Android.

Do we really gain anything with that viewModel if all it is doing is delegating to the function defined in the type then? I'm fine with having this viewModel, it just feels like we are forcing that viewModel to be there with no actual reason? Is it better to have that `@ApplicationContext` annotation rather than using `LocalContext.current` for example?

Basler182:
In my opinion, it could be implemented relatively identically: 
```kotlin
interface ContactOptionType {
    val title: String
    fun handleAction(context: Context)

    companion object {
        fun call(title: String, phoneNumber: String) = CallOption(title, phoneNumber)
    }
}

class CallOption(override val title: String, private val phoneNumber: String) : ContactOptionType {
    override fun handleAction(context: Context) {
        val dialIntent = Intent(Intent.ACTION_DIAL).apply {
            data = Uri.parse("tel:$phoneNumber")
        }
        context.startActivity(dialIntent)
    }
}

```

And use like:

```kotlin
val contactOptionType = ContactOptionType.call("Call", "1234567890")
```

 I understand, of course, that in this simple moduleit is less important to split the MVVM pattern as in larger modules. I think it would have the advantage of making the modules more standardised, e.g. not requireing to think about it when creating new modules/views.. I just wanted to share the reasons that played a role in thinking about doing it this way: things like testability, lifecycle awareness and separation of concerns.  But if there are reasons against it, I obviously have no problem with the viewmodel being left out.

pauljohanneskraft:
Perfect - sounds good - I will adapt the implementation and provide an equivalent PR for iOS 😊

Basler182:
> Perfect - sounds good - I will adapt the implementation and provide an equivalent PR for iOS 😊 

You can also wait until tomorrow after the meeting. Paul or eldi may have other views. As I'm the person who developed it, I'm probably also partly biased. 😅 I would be happy with any of those solutions.

pauljohanneskraft:
=> ```kotlin
> interface ContactOptionType {
>     val title: String
>     fun handleAction(context: Context)
> 
>     companion object {
>         fun call(title: String, phoneNumber: String) = CallOption(title, phoneNumber)
>     }
> }
> 
> class CallOption(override val title: String, private val phoneNumber: String) : ContactOptionType {
>     override fun handleAction(context: Context) {
>         val dialIntent = Intent(Intent.ACTION_DIAL).apply {
>             data = Uri.parse("tel:$phoneNumber")
>         }
>         context.startActivity(dialIntent)
>     }
> }
> ```

I implemented it similarly, hope that's fine! Since ContactOption has quite a bit more properties, I decided to only build an interface around the interface itself and keep the ContactOption data class. With that, we can then have factory functions on the companion object of ContactOption, which pretty much provides the same API as before while making the action not injected as a lambda, but rather using dynamic dispatch, as requested. What do you think?

I didn't manage to get the ViewModel with the Hilt annotations working, since it seems to not be a dependency of that module yet and it was previously also injected differently. Maybe we can discuss this in our meeting on Thursday?

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactsList.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import edu.stanford.spezi.modules.contact.model.Contact
+
+@Composable
+fun ContactsList(contacts: List<Contact>) {
+    val scrollState = rememberScrollState()
+    Column(modifier = Modifier.verticalScroll(scrollState)) {
```
eldcn, line 13: Let's use `LazyColumn` which is scrollable by default and takes care of efficiently reusing the items - similar to `LazyVStack`. Also here, let's add the modifier parameter

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
```diff
@@ -40,24 +39,26 @@ fun NavigationCard(address: String, publisher: (OnAction) -> Unit) {
...
             )
+            val context = LocalContext.current
             IconButton(
                 onClick = {
-                    publisher(OnAction.NavigateTo(address = address))
+                    val addressQuery = URLEncoder.encode(addressText, "utf-8")
+                    val gmmIntentUri = Uri.parse("geo:0,0?q=$addressQuery")
+                    val mapIntent = Intent(Intent.ACTION_VIEW, gmmIntentUri)
+                    mapIntent.setPackage("com.google.android.apps.maps")
+                    context.startActivity(mapIntent)
```
eldcn, line 55: consider wrapping this inside of a `runCatching` as it can throw in case of maps app is not installed in the device

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
```diff
@@ -40,24 +39,26 @@ fun NavigationCard(address: String, publisher: (OnAction) -> Unit) {
                 .height(IntrinsicSize.Min),
             verticalAlignment = Alignment.CenterVertically
         ) {
-            Icon(
-                Icons.Default.Place,
-                contentDescription = "Address",
-            )
+            val addressText = address.formatted()
```
eldcn, line 42: Also here, consider using remember 👍

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/AddressFormatting.kt
```diff
@@ -0,0 +1,12 @@
+package edu.stanford.spezi.modules.contact.model
+
+import android.location.Address
+
+fun Address.formatted(): String {
```
eldcn, line 5: Since we use this inside compose only, we can mark it internal and apply remember here directly:

```kotlin
@Composable
internal fun Address.formatted(): String {
    return remember(this) {
        val lines = (0..maxAddressLineIndex).map { getAddressLine(it) }
        val areaLine = listOf(locality, adminArea, postalCode).mapNotNull { it }.joinToString(" ")
        val countryLine = countryName ?: ""
        ((lines + areaLine) + countryLine)
            .filter { it.isNotBlank() }
            .joinToString("\n")
    }
}
```

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/Contact.kt
```diff
@@ -1,59 +1,30 @@
...
-    val id: UUID,
-    val icon: ImageVector?,
-    var name: String,
-    val title: String,
-    val description: String,
-    val organization: String,
-    val address: String,
+    val id: UUID = UUID.randomUUID(),
+    var name: PersonNameComponents,
+    val image: ImageVector?,
```
eldcn, line 24: Should we use new type to enable also local resources?

```kotlin
sealed interface ImageResource {
    data class Vector(val image: ImageVector) : ImageResource
    data class Drawable(@DrawableRes val resId: Int) : ImageResource
}
```

and depending on the type we use `Icon(image)` or `Icon(painter = painterResource(resId)`. If you like the approach, we can declare the new type in `design` module and create a custom icon composable taking that new type as an input. I see it to be used also in card options. Also, we can extend the data later on in case we want to support supplying an image url to be loaded asynchronously, e.g. via [coil](https://coil-kt.github.io/coil/compose/)

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/ContactOption.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * ContactOption data class used to represent a contact option.
+ *
+ * @param id the unique identifier of the contact option
+ * @param image the image of the contact option
+ * @param title the title of the contact option
+ * @param action the action of the contact option
+ */
+data class ContactOption(
+    val id: UUID = UUID.randomUUID(),
```
eldcn, line 15: Minor: I have already create a function `UUID()` for this

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/ContactOption.kt
```diff
@@ -0,0 +1,21 @@
...
+    val image: ImageVector?,
+    val title: String,
```
eldcn, line 17: Following the image comment, we have already introduced a similar type to handle strings used in compose, have a look at `StringResource` and it's text method usage in compose

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/PersonNameComponents.kt
```diff
@@ -0,0 +1,22 @@
...
+
+data class PersonNameComponents(
+    val namePrefix: String? = null,
+    val givenName: String? = null,
+    val middleName: String? = null,
+    val familyName: String? = null,
+    val nameSuffix: String? = null,
+    val nickname: String? = null,
+) {
+    fun formatted(): String {
```
eldcn, line 11: Similary as the address formatted extension function, we can mark this with composable and use this as a key for remember

eldcn (commented):
Looks very good already 👍

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactView.kt
```diff
@@ -14,55 +15,52 @@ import androidx.compose.foundation.layout.size
...
- * @see edu.stanford.spezi.modules.contact.repository.ContactRepository
+ * @see ContactOptionButton
+ * @see AddressButton
  */
 @OptIn(ExperimentalLayoutApi::class)
 @Composable
-fun ContactScreen(viewModel: ContactViewModel) {
-    val contact by viewModel.contact.collectAsState()
-
+fun ContactView(contact: Contact) {
```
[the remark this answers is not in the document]
pauljohanneskraft: I changed the order of modifier and contact to make it possible to create the ContactComposable with only specifying the contact and without parameter names, otherwise I like those changes!

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactView.kt
```diff
@@ -78,51 +76,48 @@ fun ContactScreen(viewModel: ContactViewModel) {
...
                             Column {
-                                contact?.let {
+                                Text(
+                                    text = contact.name.formatted(),
+                                    style = TextStyles.titleLarge
+                                )
+                                val subtitle = listOf(contact.title, contact.organization)
+                                    .mapNotNull { it }
+                                    .joinToString(" - ")
+                                if (subtitle.isNotBlank()) {
```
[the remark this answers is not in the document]
pauljohanneskraft: sounds good to me, done!

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactsList.kt
pauljohanneskraft: done

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactView.kt
```diff
@@ -78,51 +76,48 @@ fun ContactScreen(viewModel: ContactViewModel) {
...
                             modifier = Modifier
                                 .fillMaxWidth(),
                             horizontalArrangement = Arrangement.SpaceEvenly
                         ) {
-                            val context = LocalContext.current
-                            contact?.options?.forEach { option ->
-                                ContactOptionCard(option = option) {
-                                    viewModel.handleAction(it, context)
-                                }
+                            contact.options.forEach { option ->
```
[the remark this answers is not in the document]
eldcn: Fine by me 👍 feel free to resolve an create an issue describing your improvements

pauljohanneskraft:
> > ```kotlin
> > interface ContactOptionType {
> >     val title: String
> >     fun handleAction(context: Context)
> > 
> >     companion object {
> >         fun call(title: String, phoneNumber: String) = CallOption(title, phoneNumber)
> >     }
> > }
> > 
> > class CallOption(override val title: String, private val phoneNumber: String) : ContactOptionType {
> >     override fun handleAction(context: Context) {
> >         val dialIntent = Intent(Intent.ACTION_DIAL).apply {
> >             data = Uri.parse("tel:$phoneNumber")
> >         }
> >         context.startActivity(dialIntent)
> >     }
> > }
> > ```
> 
> I implemented it similarly, hope that's fine! Since ContactOption has quite a bit more properties, I decided to only build an interface around the action itself and keep the ContactOption data class. With that, we can then have factory functions on the companion object of ContactOption, which pretty much provides the same API as before while making the action not injected as a lambda, but rather using dynamic dispatch, as requested. What do you think?


Sorry, I have thought about this again and I think it doesn't really change much actually, if we use the dynamic dispatch approach or not. `Contact` wouldn't usually be created inside the Composable anyways and would need to be inside a `remember` anyways, so we don't really have the issue of regenerating the lambda anyways - let me know, if you agree with this or not!

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/Contact.kt
pauljohanneskraft: I like the approach, yes! I'm not so sure about the Async part, since it involves more complexity (error handling, progress views, etc), but in general having some kind of union type for images is fine - I'm wondering whether this already exists in Jetpack Compose? Like `Image` in SwiftUI also allows to create images from multiple different origins (UIImage, asset names, system names, etc).

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/AddressFormatting.kt
pauljohanneskraft: I would keep the function as is, since I would like to expose it and there it may be used in other contexts, but I will make sure to use the `remember` mechanism whenever I use it, does that sound good? also for the other case

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/Contact.kt
pauljohanneskraft: this type doesn't exist yet, right?

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/PersonNameComponents.kt
pauljohanneskraft: see other comment

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
pauljohanneskraft: Makes sense - I'd assume to at least log when an error occurs - can you guide me through the process of how to do this in this context? I know, I could just use `println`, but I'm assuming you use some other logging mechanism?

modules/contact/src/androidTest/kotlin/edu/stanford/spezi/modules/contact/ContactViewTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    fun contactView_displaysContactName() {
+        val contact = ContactFactory.mock
+
+        composeTestRule.setContent {
+            ContactView(contact)
+        }
+
+        composeTestRule.onNodeWithText(contact.name.formatted()).assertExists()
```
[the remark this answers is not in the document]
pauljohanneskraft: I have rewritten the tests to reflect this pattern.

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactView.kt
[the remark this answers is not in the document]
pauljohanneskraft: See #115

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
pauljohanneskraft: I used `println` for now - let me know what you prefer here instead.

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
```diff
@@ -66,9 +72,12 @@ fun NavigationCard(address: String, publisher: (OnAction) -> Unit) {
 
 @Composable
 @Preview
```
Basler182, line 74: Just a small project-specific thing. We have created the annotation `@ThemePreviews`, which always adds a preview of both dark and light mode. 

And if we only display small components in the preview and the whole screen is possibly obstructive, you can set the following so that you only see the component in the preview:
`SpeziTheme(isPreview = true) {`

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/Contact.kt
eldcn: nope, there is no type yet and it should be created. and no, no compose component. Regarding the async part complexity that you mentioned a new custom type can also support that, e.g. `data class Async(val url: String, fallbackImage, loading, and so on)`, basically based on the use case we can think of custom needed configuration.

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
eldcn: yeah exactly, we already have a logging mechanism, see https://github.com/StanfordSpezi/SpeziKt/tree/main/core/logging, that module is added by default via the convention plugin as a dependency to every module, so feel free to use directly

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/AddressFormatting.kt
eldcn: Sounds good to me, but let me also share my opinion why i still think it's better given the current usages to use composable. I tend to postpone planning for other usages until whenever those requirements will come.

Your other context currently is testing. There you won't have composable context, hence you wouldn't be able to use this function. Which is still fine, for the test we also want to test this function as well. If you use `formatted()` in compose and `formatted()` in the tests, we won't be able to catch if this implementation would be changed by mistake. In the test I would propose to use the expected value instead

```kotlin
fun assertHasAddress(value: Address?) {
        value?.let {
            val expectedAddressText = "build here the text from the address as done in formatted function. This way we are also testing the "
            address
                .assertExists()
                .onChildAt(0)
                .assertTextEquals(expectedAddressText)
        } ?: address.assertDoesNotExist()
    }
```

Ideally that formatted function should be unit tested - that kind of logic, it is advisable to be handled in a Mapper, which prepares the ui model to be consumed in compose. But since we are only consuming the contact in compose, I think it is still fine to test it via an android test. 

-----

No need for further adjustments for now, but just FYI regarding the simulator approach so that we won't have to repeat building the expected address text in every usage. If we would be using the address card in some other context / composable, we could extract it in a new simulator:

```kotlin
class AddressCardSimulator(
    composeTestRule: ComposeTestRule,
) {
    // this is now an identifier declared in address card
    private val address = composeTestRule.onNodeWithIdentifier(AddressCardTestIdentifier.ADDRESS)

    fun assertHasAddress(value: Address?) {
        value?.let {
            val expectedAddressText = "build here the text from the address as done in formatted function. This way we are also testing the "
            address
                .assertExists()
                .onChildAt(0)
                .assertTextEquals(expectedAddressText)
        } ?: address.assertDoesNotExist()
    }
}
```

And use it inside `ContactComposableSimulator` as follows:

```kotlin
class ContactComposableSimulator(
    private val composeTestRule: ComposeTestRule,
) {
    
    // other nodes and functions...
    
    private val address = AddressCardSimulator(composeTestRule)
    
    fun addressCard(block: AddressCardSimulator.() -> Unit) {
        address.apply(block)
    }
}
```

And contact composable test would look like this

```kotlin
@Test
fun `test displays contact address`() {
    contactComposable {
        addressCard {
            assertHasAddress(contact.address)
        }
    }
}
```

eldcn (approved):
LGTM, well done and thanks for applying the suggestions. I would just propose to use our [logger](https://github.com/StanfordSpezi/SpeziKt/tree/main/core/logging) instead of print

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/component/AddressCard.kt
pauljohanneskraft: I changed the views to make use of this now - thank you!

pauljohanneskraft:
> LGTM, well done and thanks for applying the suggestions. I would just propose to use our [logger](https://github.com/StanfordSpezi/SpeziKt/tree/main/core/logging) instead of print

I tried to incorporate this and added a logger into the companion object of `ContactOption`, not sure if this is the intended approach. Further, I wasn't sure how to use that logger in a non-class context, e.g. for AddressCard. If you know how, please let me know!

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/model/Contact.kt
pauljohanneskraft: I think this makes most sense to handle when addressing SpeziViews and I will implement this then. Adding a ticket for now, hope that's fine.

pauljohanneskraft: https://github.com/StanfordSpezi/SpeziKt/issues/124

Issue #89, opened 2024-08-18
Implement sign-out functionality


Basler182:
### Problem

When a user signs out, the app currently doesn't do anything. 

### Solution

The sign-out functionality should not only involve signing out from Firebase but also clear the navigation back stack, pop to the initial onboarding screen, and set MainUiState.Content(Routes.OnboardingScreen) in MainActivityViewModel. This will ensure the user is returned to the correct starting point after signing out, preventing any issues with the navigation stack.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #105, opened 2024-09-20
Refactor Registration Flow to Remove Consent Screen and Anonymous Login and move Invitation Code screen

Basler182:
# *Refactor Registration Flow to Remove Consent Screen and Anonymous Login and move Invitation Code screen*

- 

## :gear: Release Notes 

- Remove the consent screen and anonymous login from the registration process.
- Allow users to fully register without an invitation code.
- After login, prompt users to enter an invitation code if not already provided.
- Introduce two post-login states:
- Logged in without invitation code: User sees only the invitation code screen.
- Logged in with invitation code: User proceeds as usual.
- The user's Firestore account object will be checked for an existing invitation code to determine the user state
- When registering, only email and password are required or Google O Auth


Basler182:
Here I think it makes sense, if you have time, to have a look at it Paul and give feedback on whether the flow is now adapted as you imagined. As you know best how it should be, @pauljohanneskraft :)

pauljohanneskraft (changes requested):
Thank you for implementing this! Sorry for not going into a lot of detail, if you have questions, let me know and we can schedule a call.

Here are some points:
- We never want to sign in anonymously, it will be disabled and won't be possible. For the case of using Google authentication, we want to simply sign in with that credential, which should be possible without the go-around by anonymously signing in and then linking credentials.
- New users will sign up using the `createUser` method and then call the `enrollUser` function - you still reference `checkInvitationCode` which doesn't exist anymore. Since the server is already merged, you may want to check that behavior with the Firebase emulators - if you do not run `npm run prepare` before calling `npm run serve:seeded`, the git commit change has little to no effect, so please make sure to call `npm run prepare` whenever you check out a different commit.
- You will need to make sure to ask for the invitation code in three scenarios: 
    1. The user has not completed onboarding a single time, then we just always start from the beginning, even if the user finished some previous step. The sign in/sign up screen on iOS would then show the logged in account instead of the sign in form when an account is already signed in. The invitation code is requested afterwards depending on whether the user has already enrolled or not (by checking whether the user object exists or whether it contains an `invitationCode` property, which is pretty much the same thing at the moment).
    2. The user has finished onboarding, but is signed in with an account that is not enrolled yet. This happens, when the user finished onboarding with one account, then singed up with a new account that is not enrolled. We would need to show the sign in/sign up screen with the signed in account visible and on tap of the "next" button, we ask for the invitation code.
    3. The user has completed onboarding, but is not signed into an account. This happens, when the user finished onboarding with one account and then decided to log out. In this case, we show the sign in/sign up screen with no account signed in.

On a quick test, I initially managed to finish the onboarding by signing into an existing account, but then on a restart of the app was asked for an invitation code

Basler182:
Many thanks for the feedback @pauljohanneskraft 

> * We never want to sign in anonymously, it will be disabled and won't be possible. For the case of using Google authentication, we want to simply sign in with that credential, which should be possible without the go-around by anonymously signing in and then linking credentials.

I have adjusted it so that we no longer link the account to an anonymous account, but register it directly.

> * New users will sign up using the `createUser` method and then call the `enrollUser` function - you still reference `checkInvitationCode` which doesn't exist anymore. Since the server is already merged, you may want to check that behavior with the Firebase emulators - if you do not run `npm run prepare` before calling `npm run serve:seeded`, the git commit change has little to no effect, so please make sure to call `npm run prepare` whenever you check out a different commit.

Thanks for updating the documentation. I am now using `enrollUser` function mentioned there.  Unfortunately the emulator still doesn't work on Windows. `npm run prepare` fails. I had replaced ` ' -> \" ` as you suggested, but there are still syntactical errors. The previous install command still doesn't work either. Would it possibly make sense to containerise it as it uses platform specific stuff? Otherwise we can take another look at it together?

> * You will need to make sure to ask for the invitation code in (...)

Now whenever a user logs in or registers and is not yet enrolled, he ends up on the invitation code screen. 

> On a quick test, I initially managed to finish the onboarding by signing into an existing account, but then on a restart of the app was asked for an invitation code

It should now work as planned. Unfortunately, I could not test it extensively without the emulator. It would be great if you could check this again.

Basler182:
please re-check if there are any other things that need to be changed, or if it fits so far, @pauljohanneskraft

app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivityViewModel.kt
```diff
@@ -36,8 +37,20 @@ class MainActivityViewModel @Inject constructor(
         viewModelScope.launch {
             accountEvents.events.collect { event ->
                 when (event) {
-                    is AccountEvents.Event.SignUpSuccess, AccountEvents.Event.SignInSuccess -> {
-                        navigator.navigateTo(AppNavigationEvent.AppScreen)
+                    is AccountEvents.Event.SignInSuccess, AccountEvents.Event.SignUpSuccess -> {
+                        when (val userState = userSessionManager.getUserState()) {
```
eldcn, line 41: minor as we can extract the the common part:

```kotlin
val navigationEvent = when (val userState = userSessionManager.getUserState()) {
    UserState.NotInitialized -> {
        OnboardingNavigationEvent.OnboardingScreen
    }

    is UserState.Registered -> {
        if (userState.hasInvitationCodeConfirmed) {
            AppNavigationEvent.AppScreen
        } else {
            OnboardingNavigationEvent.InvitationCodeScreen
        }
    }
}
navigator.navigateTo(navigationEvent)
```

It would be great to also add a unit test for the `NotInitialized` user state even though that should never be the case actually.

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/login/LoginScreen.kt
```diff
@@ -169,11 +175,11 @@ You may login to your existing account or create a new one if you don't have one
...
         )
         Spacer(modifier = Modifier.height(Spacings.medium))
-        val passwordSignInOrSignUp = Action.Async.PasswordSignInOrSignUp
+        val passwordSignIn = Action.Async.PasswordSignIn
         AsyncTextButton(
-            isLoading = uiState.pendingActions.contains(passwordSignInOrSignUp),
-            text = if (uiState.isAlreadyRegistered) "Login" else "Register",
-            onClick = { onAction(passwordSignInOrSignUp) },
+            isLoading = uiState.pendingActions.contains(passwordSignIn),
+            text = "Login",
```
eldcn, line 181: Let's declare this as well in string resources

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/login/LoginViewModel.kt
```diff
@@ -64,30 +59,26 @@ internal class LoginViewModel @Inject constructor(
...
+        if (validator.isFormValid(uiState)) {
             authenticationManager.signIn(
                 email = email,
                 password = password,
             ).onSuccess {
                 accountEvents.emit(event = AccountEvents.Event.SignInSuccess)
             }.onFailure {
                 accountEvents.emit(event = AccountEvents.Event.SignInFailure)
-                messageNotifier.notify("Failed to sign in")
+                messageNotifier.notify(R.string.error_sign_in_failed)
```
eldcn, line 78: 👏

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
```diff
@@ -79,11 +82,24 @@ internal class UserSessionManagerImpl @Inject constructor(
         )
     }
 
+    override suspend fun forceRefresh() {
```
eldcn, line 85: can we wrapp this in a runCatching block?

modules/account/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,28 @@
...
+    <string name="cancel">Cancel</string>
+    <string name="create_a_new_account">Create a new Account</string>
+    <string name="please_fill_out_the_details_below_to_create_your_new_account">Please fill out the details below to create your new account.</string>
+    <string name="e_mail_address">E-Mail Address</string>
+    <string name="first_name">First Name</string>
+    <string name="last_name">Last Name</string>
+    <string name="gender">Gender</string>
+    <string name="select_gender">Select Gender</string>
+    <string name="date_of_birth">Date of Birth</string>
+    <string name="select_date">Select Date</string>
```
eldcn, line 26: we have removed these from the registration screen (name, gender and birthday). Should we remove the strings as well as they are not used?

modules/account/src/test/java/edu/stanford/spezi/module/account/manager/UserSessionManagerTest.kt
```diff
@@ -262,6 +257,25 @@ class UserSessionManagerTest {
...
 
+    private fun setupHasConfirmedInvitationCode() {
+        val uid = "uid"
+        val document: DocumentSnapshot = mockk {
+            every { getString("invitationCode") } returns "1234"
+        }
+        val firebaseUser: FirebaseUser = mockk {
+            every { isAnonymous } returns false
+        }
+        val documentTask: Task<DocumentSnapshot> = mockk {
```
eldcn, line 268: I have created a `mockTask` function, feel free to use that here. If you can replace other usages as well would be great, just search for ` every { isComplete } returns true`

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/register/RegisterFormValidator.kt
```diff
@@ -1,44 +1,14 @@
 package edu.stanford.spezi.module.account.register
 
-import java.time.LocalDate
 import javax.inject.Inject
 
 internal class RegisterFormValidator @Inject constructor() : FormValidator() {
```
eldcn, line 5: this and `LoginFormValidator` aren't needed anymore, we can make `FormValidator` injectable and use that directly - maybe rename to `CredentialsValidator` or `AuthValidator`

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
```diff
@@ -79,11 +82,24 @@ internal class UserSessionManagerImpl @Inject constructor(
...
+        firebaseAuth.currentUser?.getIdToken(true)?.await()
+    }
+
+    @Suppress("UnusedPrivateMember")
```
eldcn, line 89: Is there any plan to add the consent again?

Other question: What should happen with other related components: ConsentScreen, related routes and navigation events and the pdf creation service? All these are only used in tests now. Should we at least remove the consent screen from the nav graph in main activity? If so we should adjust the app navigation tests as well

eldcn (approved):
LGTM, well done 👏 Minor comments only

modules/account/src/main/res/values/strings.xml
Basler182: good idea, i had forgotten about the string resources

modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/UserSessionManager.kt
Basler182: consent screen is also part of the spezi framework, so it will still be there, even if the app itself does not currently use it. the code for this should therefore remain in place for the time being. removing the consent screen from nav graph is a good idea.

pauljohanneskraft (approved):
LGTM! One minor adaption would be to allow for going back after signing up but before entering the invitation code, but I feel like it may be more fitting with https://github.com/StanfordSpezi/SpeziKt/issues/89

PR #110, opened 2024-09-23
BLE Devices persistence and pairing mode

### LINKED ISSUE #51: Improve: Local Storage
### Problem

Currently, our components are not written in a way that makes unit testing feasible, at least partly. This lack of testability complicates the validation of our storage mechanisms and ensures robust functionality. Specifically, we need to address the issue of creating and managing encrypted files and standard files within our local storage system, which are not easily unit-testable in their current form.

### Solution

To improve testability, we propose the creation of a FileHandler or similar component with methods such as createEncryptedFile, createFile, deleteFile, and exists(filename). This component should be injected into EncryptedFileStorage. While FileHandler will be non-unit-testable, it will enable unit testing of EncryptedFileStorage by allowing the injection of a mocked FileHandler and subsequent stubbing and verification of expected behavior. Additionally, we can add androidTests for FileHandler or exclude it from the code coverage in the build.gradle project file under the setupJacoco method.

Furthermore, we should mark  `sealed class PreferenceKey<T>(val key: Preferences.Key<T>) {` key
 as internal val key to prevent consumers of this module from needing to add a datastore dependency.

Additionally, we should implement a singleton KeyValueStorageFactory component with a method createStorage(name: String, type: (DataStore/SharePrefs): KeyValueStorage. The factory will maintain a Map of created storages, returning an existing storage if one with the same name and type already exists, or creating a new one if not. This approach ensures that no more than one datastore with the same name is created.

The LocalKeyValueStorage and EncryptedSharedPreferencesKeyValueStorage can than be marked as internal and should not require an @Inject constructor, as they will be created within the factory. The extension property datastore on the companion object can be moved to the class level. The factory will prefix the filenames with edu.stanford.spezi, allowing for organized and grouped files in the device explorer.

### Additional context

_No response_

### Code of Conduct

### LINKED ISSUE #99: BottomSheet not reopening after being closed unless content changes
### Description

When interacting with a UI element (e.g., Message Vitals) that should open a BottomSheet, closing the BottomSheet and then trying to reopen it does not work. The BottomSheet only reappears if its content changes. This issue seems to be caused by the LaunchedEffect in the AppScreen.

### Reproduction

1. Open a UI interaction that triggers a BottomSheet (e.g., Message Vitals).
2. Close the BottomSheet.
3. Try to reopen the same BottomSheet.
4. The BottomSheet does not appear again unless its content changes.

### Expected behavior

The BottomSheet should reopen when the same UI interaction is triggered, regardless of whether the content has changed.

### Additional context

This issue appears to be related to how LaunchedEffect is handled in the AppScreen component.

### Code of Conduct

eldcn:
# BLE Devices persistence and pairing mode

## What was done
- Fixes #51 
- Fixes #99 where the bottom sheet can't reopen the same content
- Improve local storage API to support complex type persistence as json string
- Add support in bluetooth module to listen to pairing device changes
- Persist paired devices with additional metadata such as connected state and last seen timestamp
- Add pairing mode bottom sheet
- Unit test new components and adjust existing tests

**Recording**

https://github.com/user-attachments/assets/5dd997ef-4294-45af-89bb-c97a9ac9a343

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing


core/bluetooth/src/main/kotlin/edu/stanford/spezi/core/bluetooth/domain/BLEPairedDevicesStorage.kt
```diff
@@ -0,0 +1,125 @@
...
+            name = device.name,
+            connected = connected,
+            lastSeenTimeStamp = timeProvider.currentTimeMillis()
+        )
+
+        update(devices = currentDevices + newDevice)
+    }
+
+    private fun refreshState() = execute {
+        val systemBoundDevices = bluetoothAdapter.bondedDevices
```
Basler182, line 58: `bluetoothAdapter.bondedDevices` can return null depending on the state of the Bluetooth adapter.

```suggestion
        val systemBoundDevices = bluetoothAdapter.bondedDevices ?: return@execute
```

Basler182 (approved):
Great work on fixing multiple issues! I also tested the pairing process with my phone, and everything worked smoothly. Overall, really good job - just a small comment to follow up on.

core/bluetooth/src/main/kotlin/edu/stanford/spezi/core/bluetooth/domain/BLEPairedDevicesStorage.kt
eldcn: Thanks 🙌 commited your suggestion. But in general execute launches the io scope which already catches the exceptions via the coroutine exception handler, this for instance can also throw in case of missing permission, hence it is safe to use execute

Basler182: Thanks for the explanation, good to know! 👍

PR #113, opened 2024-09-28
added secrets

Basler182:
# *added secrets*

file was not injected at build time

## :gear: Release Notes 
file gets written to folder at build time

while working on this PR I noticed that this file should generally not be here in the account module. during the framework optimisations of the account module, I think we should move on to creating an interface, e.g. ServerClientIdProvider and use that to get the id in the module, so that you can provide the implementation from outside of the module or possibly think about further possible mechanisms


modules/account/src/main/kotlin/edu/stanford/spezi/module/account/manager/AuthenticationManager.kt
```diff
@@ -137,8 +138,7 @@ internal class AuthenticationManager @Inject constructor(
             val googleIdOption = GetGoogleIdOption.Builder()
                 .setFilterByAuthorizedAccounts(filterByAuthorizedAccounts)
                 .setAutoSelectEnabled(true)
-                // TODO: Uncomment once secrets xml has been added in CI secrets
-                // .setServerClientId(context.getString(R.string.serverClientId))
+                .setServerClientId(context.getString(R.string.serverClientId))
```
eldcn, line 141: this won't compile locally as we don't have this resource, you need to add the `secrets.xml` file nonetheless with a dummy / empty string. you can also add a comment that get's overriden with the actual id in CI

.github/workflows/main.yml
```diff
@@ -42,8 +42,12 @@ jobs:
           bundler-cache: true
       - name: Decode and Write Google Service JSON
         env:
-            GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
+          GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
         run: echo $GOOGLE_SERVICES_JSON | base64 --decode >./app/google-services.json
+      - name: Decode and Write Secrets
+        env:
+          SERVER_CLIENT_ID: ${{ secrets.SERVER_CLIENT_ID }}
+        run: echo $SERVER_CLIENT_ID | base64 --decode >./modules/account/src/main/res/values/secrets.xml
```
PSchmiedmayer, line 50: We might want to rename this to `SECRETS_XML`
```suggestion
      - name: Decode and Write Secrets
        env:
          SECRETS_XML: ${{ secrets. SECRETS_XML }}
        run: echo $SECRETS_XML | base64 --decode >./modules/account/src/main/res/values/secrets.xml
```

README.md
```diff
@@ -22,16 +22,35 @@ Spezi is a collection of modules that can be used to build Android applications
...
+### Continuous Integration and Delivery Setup
 
 #### Google Play Internal Deployment
 
 First, create a Google Cloud Services Account and corresponding JSON secrets key in accordance to the [fastlane supply](https://docs.fastlane.tools/actions/supply/) documentation. Store the JSON representation of the key in a `SERVICE_ACCOUNT_JSON_KEY` secret available to the GitHub action.
 
+Follow along
+the [Set up your Google APIs console](https://developer.android.com/identity/sign-in/credential-manager-siwg#set-google)
+documentation to create a OAuth client ID. Store secrets.xml representation of the key in
+a `SERVER_CLIENT_ID` secret available to the GitHub action.
```
PSchmiedmayer, line 34:
```suggestion
a `SECRETS_XML` secret available to the GitHub action.
```

PSchmiedmayer (approved):

PR #117, opened 2024-10-04
fixed to correctly get the ZoneOffset

Basler182:
# *fixed to correctly get the ZoneOffset*

The DateTime of the data records was displayed incorrectly in the views because the time zone was not correctly taken into account.

## :gear: Release Notes 
The problem is solved by reading the time zone correctly. However, the PR does not take into account that the offset of the record is ignored when converting records to observations. Instead, the default time zone of the device is used, which is correct in most cases, including ours.


modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/ObservationsDocumentMapperImpl.kt
```diff
@@ -104,8 +102,11 @@ class ObservationsDocumentMapperImpl @Inject constructor(
         )
 
         return WeightRecord(
-            time = time,
-            zoneOffset = zoneOffset,
+            time = observation.effectiveDateTimeType.value.toInstant(),
+            zoneOffset = ZoneOffset.ofHoursMinutes(
```
eldcn, line 106: Would it make sense to extract this in a separate function or extension function and reuse throughout this class?

```kotlin
private fun getZoneOffset(observation: Observation) = ZoneOffset.ofHoursMinutes(
    observation.effectiveDateTimeType.tzHour,
    observation.effectiveDateTimeType.tzMin
)
```

Additionally, I see that `observation.effectiveDateTimeType` can throw if `effective` property of observation is not a `DateTimeType`. This can happen in `RecordToObservationMapperImpl` in the `T.createObservation` function in case start and end Date is not equal. QQ: I know that it was never an issue so far, but do you think that we need to safely access `effectiveDateTimeType` property and handle errors as well?

eldcn (approved):
LGTM, just a small proposal and a question to clarify 👍

modules/healthconnectonfhir/src/main/java/edu/stanford/healthconnectonfhir/ObservationsDocumentMapperImpl.kt
Basler182: This should not be a problem, as we only save values that correspond to the basic condition. However, additional security can't hurt. How would we make the adjustment? Presumably the `map` function of the `ObservationsDocumentMapper` would then return a `result`, correct?

eldcn: I would just safely handle the offset, no need to fail the whole mapping due to offset. Maybe just using utc?

```kotlin
private fun getZoneOffset(observation: Observation) = runCatching {
    ZoneOffset.ofHoursMinutes(
        observation.effectiveDateTimeType.tzHour,
        observation.effectiveDateTimeType.tzMin
    )
}.getOrDefault(ZoneOffset.UTC)
```

Basler182: It might actually make also sense to use ZoneOffset.systemDefault() instead of UTC in most cases, as this would probably be more correct. But if we don't have any information about the offset, it might be technically better to use .getOrNull, since ZoneOffset is nullable in the record and would still be fine. However, the time information is not nullable, and if we don't have the offset, the time is missing as well. Without the time specification, the health data loses its temporal reference and therefore its value. In this case, it might even be correct that the mapping fails.

eldcn: I see, then I would say we can just use system default. It's just an edge case and it should be fine imo 👍

PR #118, opened 2024-10-04
Feature/sign out #89

Basler182:
# *Feature/sign out*

When a user signs out, the app currently doesn't do anything.

## :gear: Release Notes 
- User is returned to the correct starting point after signing out.

### DIFF
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
@@ -215,7 +215,7 @@ class MainActivity : FragmentActivity() {
 
                         is AppNavigationEvent.AppScreen -> navHostController.navigateTo(
                             route = Routes.AppScreen,
-                            clearBackStack = true
+                            clearBackStack = event.clearStackTrace
                         )
 
                         is NavigationEvent.PopBackStack -> navHostController.popBackStack()
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
@@ -201,8 +201,8 @@ class MainActivity : FragmentActivity() {
                             Routes.InvitationCodeScreen
                         )
 
-                        is OnboardingNavigationEvent.OnboardingScreen -> navHostController.navigate(
-                            Routes.OnboardingScreen
+                        is OnboardingNavigationEvent.OnboardingScreen -> navHostController.navigateTo(
+                            Routes.OnboardingScreen, event.clearBackStack
                         )
 
                         is OnboardingNavigationEvent.SequentialOnboardingScreen -> navHostController.navigate(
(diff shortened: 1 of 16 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -78,4 +78,17 @@
...
+    <string name="ble_device_discovering_subtitle">Halten Sie die Bluetooth-Taste 3 Sekunden lang gedrückt, um das Gerät in den Pairing-Modus zu versetzen.</string>
+    <string name="ble_device_found_title">Zubehör koppeln</string>
+    <string name="ble_device_found_subtitle">Möchten Sie %1$s mit der ENGAGEHF-App koppeln?</string>
+    <string name="ble_device_paired_title">Zubehör Gepaart</string>
+    <string name="ble_device_paired_subtitle">%1$s wurde erfolgreich mit der ENGAGEHF-App gekoppelt</string>
+    <string name="ble_device_error_title">Fehler</string>
+    <string name="ble_device_error_subtitle">Es ist ein Fehler aufgetreten, bitte versuchen Sie es später noch einmal.</string>
+    <string name="your_devices_header_title">Deine Geräte</string>
+    <string name="last_seen_on">Zuletzt gesehen am %1$s</string>
+    <string name="sign_out_failed">Abmeldung fehlgeschlagen.</string>
```
eldcn, line 93: Danke 👏

modules/onboarding/src/main/kotlin/edu/stanford/spezi/module/onboarding/OnboardingNavigationEvent.kt
```diff
@@ -8,4 +8,5 @@ sealed class OnboardingNavigationEvent : NavigationEvent {
     data object OnboardingScreen : OnboardingNavigationEvent()
     data object SequentialOnboardingScreen : OnboardingNavigationEvent()
     data object ConsentScreen : OnboardingNavigationEvent()
+    data object ClearBackStackOnboarding : OnboardingNavigationEvent()
```
eldcn, line 11: I would propose to reuse the existing event and parametrise it:

```kotlin
data class OnboardingScreen(val clearBackStack: Boolean) : OnboardingNavigationEvent()
```

And you can emit this event with flag true on sign up. You can adapt listening the event in the activity then as follows:

```kotlin
is OnboardingNavigationEvent.OnboardingScreen -> navHostController.navigate(
    Routes.OnboardingScreen,
) {
    if (event.clearBackStack) {
        popUpTo(navHostController.graph.startDestinationId) { inclusive = true }
        launchSingleTop = true // we need this config as well
    }
}
```

app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
```diff
@@ -216,6 +216,11 @@ class MainActivity : FragmentActivity() {
                         is AppNavigationEvent.AppScreen -> navHostController.navigate(Routes.AppScreen)
```
eldcn, line 216: We need the same clearing back stack mechanism for this event as well, here we need to clear the back stack when we receive the sign in / sign up success event in the view model, e.g. `AppNavigationEvent.AppScreen(clearBackStack = true)`, and with the flag false when calculating the start destination. Currently we are able to go back after the login which we should avoid.

With this, we can create the following extension function on nav host controller

```kotlin
private fun NavHostController.navigateTo(route: Routes, clearBackStack: Boolean = false) {
    navigate(route) {
        if (clearBackStack) {
            popUpTo(graph.startDestinationId) { inclusive = true }
            launchSingleTop = true
        }
    }
}
```

And reuse for both `AppScreen` and `OnboardingScreen` event

modules/onboarding/src/main/kotlin/edu/stanford/spezi/module/onboarding/OnboardingNavigationEvent.kt
Basler182: oh yes that's right, that makes a lot more sense, I don't know why I didn't think of it

app/src/main/kotlin/edu/stanford/bdh/engagehf/MainActivity.kt
Basler182: yes, it makes total sense that we also introduce this with the AppScreen

app/src/main/kotlin/edu/stanford/bdh/engagehf/navigation/AppNavigationEvent.kt
```diff
@@ -3,6 +3,6 @@ package edu.stanford.bdh.engagehf.navigation
 import edu.stanford.spezi.core.navigation.NavigationEvent
 
 sealed interface AppNavigationEvent : NavigationEvent {
-    data object AppScreen : AppNavigationEvent
+    data class AppScreen(val clearStackTrace: Boolean) : AppNavigationEvent
```
eldcn, line 6: `clearBackStack`*

eldcn (approved):
🚀

Issue #36, opened 2024-06-11
Onboarding Views Layout Improvements


PSchmiedmayer:
### Problem

The current onboarding views all look cramped up at the top of the screen.

<div style="display: flex; justify-content: space-between;">
  <img src="https://github.com/StanfordSpezi/SpeziKt/assets/28656495/78bc2649-c9cc-4a0d-8c41-3923777787a0" alt="OnboardingFlow" width="265" />
  <img src="https://github.com/StanfordSpezi/SpeziKt/assets/28656495/b98575e0-e30d-4acb-9eef-1e59de9160fc" alt="OnboardingFlow2" width="265" />
</div>

### Solution

On iOS, we are moving more and more to utilizing a [Navigation Bar](https://developer.apple.com/design/human-interface-guidelines/navigation-bars) at the top or facilitate a decent spacing.

I would suggest to utilize a [top app bar](https://m3.material.io/components/top-app-bar/guidelines) to facilitate a common top bar across the onboarding flow (should be defined in the app, not the composable views themselves to allow future customization and the usage of the views within other part of applications).

The individual views should be reduced to a simple composable that can be embedded in a larger view hierarchy including a top navigation bar structure.
We should also ensure that we have a screen-adaptive layout that tries to fill the application's space (up to a max with and height to ensure a proper layout on tablet devices).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #122, opened 2024-10-12
Task/onboarding improvements #36

Basler182:
# *onboarding improvements*

#36 

## :gear: Release Notes 
- utilize navigation bar
- refactored code
- extracted common logic
- extracted strings into ressources


core/design/src/main/kotlin/edu/stanford/spezi/core/design/component/CommonScaffold.kt
```diff
@@ -0,0 +1,53 @@
...
+                        verticalAlignment = Alignment.CenterVertically
+                    ) {
+                        Text(text = title)
+                    }
+                },
+            )
+        },
+        content = { innerPadding ->
+            Column(modifier = Modifier.padding(innerPadding)) {
+                content(Modifier.padding(innerPadding))
```
eldcn, line 38: Should't be need to pass this modifier with the inner padding to the content imo, `Modifier.fillMaxSize().padding(innerPadding)` should be enough. Also, since you are using a `Column` here, would be good to indicate to the calllers of the CommonScaffold by updating content type to `content: @Composable ColumnScope.() -> Unit,`. Also let's add a `modifier: Modifier = Modifier` to the common scaffold and pass it internally in the scaffold

eldcn (approved):
🚀

core/design/src/main/kotlin/edu/stanford/spezi/core/design/component/CommonScaffold.kt
Basler182: good idea, i have adapted it accordingly 👍

Issue #124, opened 2024-10-15
SpeziViews: ImageResource union type


pauljohanneskraft:
### Problem

We currently use ImageVector for images, but may want to support all kinds of images later on. https://github.com/StanfordSpezi/SpeziKt/pull/104#discussion_r1781637439

### Solution

Implement ImageResource to handle all kinds of image types and possibly also a view to display them easily.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #133, opened 2024-10-26
feat union type image resource

Basler182:
# *feat union type image resource *

#124 

## :gear: Release Notes 
- added union type image resource


modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactComposable.kt
```diff
@@ -76,14 +76,13 @@ fun ContactComposable(contact: Contact, modifier: Modifier = Modifier) {
                 horizontalArrangement = Arrangement.spacedBy(Spacings.small),
                 verticalAlignment = Alignment.CenterVertically
             ) {
-                val image = contact.image ?: Icons.Default.AccountBox
```
Basler182, line 79: just a small detail, but maybe a minor imporevement. currently the image in the `data class contact` is nullable and has the default value `null`, wouldn't it be worth considering removing the nullability and  setting the default value of the `contact` directly to `ImageResource.Vector(Icons.Default.AccountBox)` and then save us the null check at this point to set a Default Value or is it planned to use the image in other ways where it is needed to be "really" null? @pauljohanneskraft

core/design/src/main/kotlin/edu/stanford/spezi/core/design/component/ImageResource.kt
```diff
@@ -0,0 +1,19 @@
...
+import edu.stanford.spezi.core.utils.UUID
+
+/**
+ * A sealed class to represent an image resource. It can be either a vector image or a drawable.
+ * This is useful to abstract the image resource type and use it in a composable function. The identifier can be used for tests.
+ * @see ImageResource.Vector
+ * @see ImageResource.Drawable
+ * @see ImageResourceIcon
+ */
+sealed class ImageResource(open val identifier: String = UUID().toString()) {
```
eldcn, line 14: This property is declared open, but none of the types is overriding it, and I also see no need to override. Feel free to move it out of constructor as a property of the sealed class itself:

```kotlin
sealed class ImageResource { 
       val identifier: String = UUID().toString()
       // other types
}
```

Even though we have one single usage for loading an image via a url (ExpandableVideoSection.kt) where we use `SubcomposeAsyncImage`, I would add a third case in this sealed class, and ideally replace the usage there as well via the model

```kotlin
data class Remote(val url: String) : ImageResource()
```

core/utils/src/main/kotlin/edu/stanford/spezi/core/utils/extensions/SemanticsPropertyKey.kt
```diff
@@ -0,0 +1,5 @@
+package edu.stanford.spezi.core.utils.extensions
+
+import androidx.compose.ui.semantics.SemanticsPropertyKey
+
+val ImageResourceKey = SemanticsPropertyKey<String>("ImageResourceKey")
```
eldcn, line 5: Let's have a full fledged setup for this 🙌 

We can add the following extension function in `Modifier.kt` in utils module:

```kotlin
fun Modifier.imageResourceIdentifier(identifier: String) = testTag(tag = identifier)
    .semantics { this[ImageResourceKey] = identifier }
```

It adds by default the tag and the image resource key. This allows replacing the usages to: `modifier = Modifier.imageResourceIdentifier(imageResource.identifier)`

Regarding testing, you can add a new file in `testing` module with this extension:

```kotlin
fun SemanticsNodeInteraction.assertImageIdentifier(identifier: String) = assert(
    SemanticsMatcher.expectValue(ImageResourceKey, identifier)
)
```

These will result in:

```kotlin
@Composable
fun ImageResourceIcon(
    imageResource: ImageResource,
    contentDescription: String,
    modifier: Modifier = Modifier,
    tint: Color = Colors.primary,
) {
    val imageModifier = modifier.then(Modifier.imageResourceIdentifier(imageResource.identifier))
    when (imageResource) {
        is ImageResource.Vector -> {
            Icon(
                imageVector = imageResource.image,
                contentDescription = contentDescription,
                tint = tint,
                modifier = imageModifier
            )
        }

        is ImageResource.Drawable -> {
            Icon(
                painter = painterResource(id = imageResource.resId),
                contentDescription = contentDescription,
                tint = tint,
                modifier = imageModifier
            )
        }
    }
}
```

And in tests you assert as follows:

```kotlin
fun assertHasImage(imageResource: ImageResource?) {
    imageResource?.let {
        image(imageResource)
            .assertExists()
            .assertContentDescriptionContains("Profile Picture")
            .assertImageIdentifier(it.identifier)
    }
}
```

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactComposable.kt
eldcn: I like the idea of providing a default value and getting rid of the nullability 🙌

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactComposable.kt
```diff
@@ -76,14 +76,13 @@ fun ContactComposable(contact: Contact, modifier: Modifier = Modifier) {
...
-                    image,
+                val image = contact.image ?: ImageResource.Vector(Icons.Default.AccountBox)
+                ImageResourceIcon(
+                    imageResource = image,
                     contentDescription = stringResource(R.string.profile_picture),
-                    tint = Colors.primary,
                     modifier = Modifier
                         .size(Sizes.Icon.medium)
-                        .testIdentifier(ContactComposableTestIdentifier.IMAGE, image.name)
+                        .testIdentifier(ContactComposableTestIdentifier.IMAGE, image.identifier)
```
eldcn, line 85: with the imageResourceIdentifier proposed above there is no need to specify the test identifier as it sets the test tag by default and we can use the same id of the image to query the node

eldcn (commented):
🚀 

It would be great if you can create an issue to replace all other icon usages with the new appoach, I will handle that in a later point

core/design/src/main/kotlin/edu/stanford/spezi/core/design/component/ImageResource.kt
eldcn: An additional comment, please also annotate the ImageResource with `@Immutable` - an additional hint to compose to avoid unnecessary recompositions because we guarantee that the instance won't change at runtime.

Basler182: Yes, that makes sense. I still had it like that because I thought it might make sense to overwrite it, for example to set the url as identifier or the resIdltoString but I think it fits to just pull it out from the constructor 👍

core/testing/src/main/kotlin/edu/stanford/spezi/core/testing/SemanticsMatcher.kt
```diff
@@ -0,0 +1,7 @@
+package edu.stanford.spezi.core.testing
+
+import androidx.compose.ui.test.SemanticsMatcher
+import edu.stanford.spezi.core.utils.extensions.ImageResourceKey
+
+fun hasIconIdentifier(expectedIdentifier: String) =
```
eldcn, line 6: can be removed now?

core/design/src/main/kotlin/edu/stanford/spezi/core/design/component/ImageResourceComposable.kt
```diff
@@ -0,0 +1,116 @@
...
+        is ImageResource.Remote -> {
+            SubcomposeAsyncImage(
+                model = imageResource.url,
+                modifier = modifier,
+                contentDescription = contentDescription,
+            ) {
+                val state = painter.state
+                val painter = painter
+                if (state is AsyncImagePainter.State.Loading) {
+                    Box(Modifier.matchParentSize()) {
```
eldcn, line 65: `ShimmerEffectBox` could also be an option here

eldcn (approved):
🚀

core/design/src/main/kotlin/edu/stanford/spezi/core/design/component/ImageResourceComposable.kt
Basler182: Yeah, good Idea. I exchanged it!

modules/contact/src/main/kotlin/edu/stanford/spezi/modules/contact/ContactComposable.kt
Basler182: @pauljohanneskraft  fyi: I have decided to remove the nullabllity 👍

PR #137, opened 2024-10-27
fix #90 don't dismiss failed message action

### LINKED ISSUE #90: Message Action Complete Handling
### Problem

At present, the message action is not based on whether it was successful, but always completed

### Solution

It should be checked whether the action is successful and only then should the message be completed.

### Additional context

_No response_

### Code of Conduct

Basler182:
# *don't dismiss failed message action*

#90 

## :gear: Release Notes 
- don't dismiss failed message action


app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -220,7 +224,15 @@ class BluetoothViewModel @Inject internal constructor(
     ) {
         when (messagesAction) {
             is MessagesAction.HealthSummaryAction -> {
-                handleHealthSummaryAction(messageId)
+                handleHealthSummaryAction(messageId).onFailure {
+                    return // on failure, we don't want to dismiss the message
```
eldcn, line 228: very minor as I tend to avoid comments and want the code to speak instead 😅 . Before when I would create a `var shouldDismissMessage = isDismissable` and in both failures here force set it to false `shouldDismissMessage = false`

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -242,32 +254,27 @@ class BluetoothViewModel @Inject internal constructor(
...
-        ).getOrNull()?.let { video ->
-            navigator.navigateTo(
-                EducationNavigationEvent.VideoSectionClicked(
-                    video = video
-                )
+        ).onSuccess { video ->
+            navigator.navigateTo(EducationNavigationEvent.VideoSectionClicked(video))
+        }.onFailure {
+            messageNotifier.notify(
+                message = "Error while handling message action."
```
eldcn, line 271: Can we extract to string resources? Message notifier supports notify(resId) already

modules/education/src/main/java/edu/stanford/spezi/modules/education/videos/EducationViewModel.kt
```diff
@@ -27,11 +29,11 @@ internal class EducationViewModel @Inject constructor(
...
         viewModelScope.launch {
             _uiState.value = UiState.Loading
-            val result = educationRepository.getVideoSections()
-            _uiState.value = if (result.isSuccess) {
-                UiState.Success(EducationUiState(videoSections = result.getOrNull() ?: emptyList()))
-            } else {
-                UiState.Error("Failed to load video sections")
+            educationRepository.getVideoSections().onFailure {
+                logger.e(it) { "Failed to load video sections" }
+                _uiState.value = UiState.Error("Failed to load video sections")
```
eldcn, line 34: also here, extracting the string to string resources and updating the error message type to a `StringResource` would be great

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/BluetoothViewModel.kt
```diff
@@ -218,9 +223,20 @@ class BluetoothViewModel @Inject internal constructor(
         messageId: String,
         isDismissible: Boolean,
     ) {
+        var shouldDismissMessage = isDismissible
         when (messagesAction) {
             is MessagesAction.HealthSummaryAction -> {
-                handleHealthSummaryAction(messageId)
+                handleHealthSummaryAction(messageId).onFailure {
+                    shouldDismissMessage = false
+                    return
```
eldcn, line 231: return is also not needed anymore imo, either return only + comment as it was before or address via the new `shouldDismissMessage` variable which will skip dismissing below

Basler182: i left the `return` in because i thought it would make it easier to read, that nothing more would happen in this case, but i removed it now

PR #145, opened 2024-12-08
fix metric conversion

Basler182:
# *fix metric conversion*

lbs isn't converted into kg. 

### DIFF
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/service/Measurement.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/service/Measurement.kt
@@ -53,7 +53,7 @@ sealed interface Measurement {
      * Represents a weight measurement.
      */
     data class Weight(
-        val weight: Double,
+        val weight: Double, // in kg
         val zonedDateTime: ZonedDateTime?,
         val userId: Int?,
         val bmi: Double?,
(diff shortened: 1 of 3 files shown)


eldcn (approved):
Thanks @Basler182 for jumping in with this 🚀 

I just tested with my device and the conversion works as expected for both units. Feel free to merge already

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/service/Measurement.kt
```diff
@@ -53,7 +53,7 @@ sealed interface Measurement {
      * Represents a weight measurement.
      */
     data class Weight(
-        val weight: Double,
+        val weight: Double, // in kg
```
eldcn, line 56: No need to address now, but I was thinking whether we should better store the corresponding value and unit measured with the device settings when the measurement occurred. I mean I used `lbs` setting in the device, but then we display it based on the locale of the phone which might not always match. We can avoid that and simply display the value and unit that we receive from the measurement instead of considering the locale. I was thinking the following:

```kotlin
data class Weight(
    val weight: Double,
    val unit: Unit,
    val zonedDateTime: ZonedDateTime?,
    val userId: Int?,
    val bmi: Double?,
    val height: Double?,
) : Measurement {
    enum class Unit {
        KG, LBS
    }
}

// in the mapper

val flags = data[0]
val unitKg = (flags and 0b00000001).toInt() == 0 // Check if bit 0 is 0 for kg, 1 for lb
val unit = if (unitKg) {
    Measurement.Weight.Unit.KG
} else {
    Measurement.Weight.Unit.LBS
}
val weight = if (unitKg) {
    // Kilograms, resolution of 0.005 kg
    (data[1].toInt() and 0xFF or (data[2].toInt() and 0xFF shl 8)) * 0.005
} else {
    // Pounds, resolution of 0.01 lb and translating ancient into metric clarity
    (data[1].toInt() and 0xFF or (data[2].toInt() and 0xFF shl 8)) * 0.01
}
```

Basler182: one could also consider whether the app should generally offer an option to switch between lbs and kg. Currently, all views automatically assume that users from the US prefer lbs. However, there may also be users there who prefer the metric system.

PR #146, opened 2024-12-08
Error handling when generating Health summary

eldcn:
# *Health summary PDF generation*

- Even though the summary pdf was generated successfully, the service was throwing because there was no PDF viewer app installed in the device
- We were able to catch the error and were attempting to show a toast. However, since the code is being executed in a background thread, the app was crashing because toast must be executed in the main thread.

Addressed the issue by:
- Ensuring that toast messages are executed always in the main thread
- Explicitly check whether after PDF generation whether there is an appropriate pdf reader app installed in the device. If not we display a toast message that summary is generated successfully but no pdf reader installed in the device. In case of success, the user will have the option to select the the pdf reader app via a default android bottom sheet

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -35,4 +35,12 @@
         </provider>
     </application>
 
+    <queries>
+        <!-- Intent query to open PDF files -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:mimeType="application/pdf" />
+        </intent>
+    </queries>
+
 </manifest>
--- a/app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/HealthSummaryService.kt
+++ b/app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/HealthSummaryService.kt
@@ -5,6 +5,7 @@ import android.content.Intent
 import android.os.Environment
 import androidx.core.content.FileProvider
 import dagger.hilt.android.qualifiers.ApplicationContext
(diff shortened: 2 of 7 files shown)


PSchmiedmayer (approved):
Thank you for fixing this @eldcn!

I am wondering if there a default PDF viewer app that we could point a user to?
Or does Chrome have that integrated so we could try to open the PDF in chrome (assuming that Chrome is installed on most Android Phones by default)?

eldcn:
> Thank you for fixing this @eldcn!
> 
> I am wondering if there a default PDF viewer app that we could point a user to? Or does Chrome have that integrated so we could try to open the PDF in chrome (assuming that Chrome is installed on most Android Phones by default)?

This is exactly what our snippet is doing. It is basically scanning all installed apps in the device, that have declared in their `AndroidManifest.xml` an activity that is capable to view pdfs, e.g.

```xml
<activity android:name=".APdfViewerActivity">
    <intent-filter>
        <action android:name="android.intent.action.VIEW" />
        
        <category android:name="android.intent.category.DEFAULT" />
        
        <data android:scheme="content" />
        <data android:scheme="file" />
        <data android:mimeType="application/pdf" />
    </intent-filter>
</activity>
```

In case there are multiple apps that have declared such an intent filter, the first time, Android will show a bottom sheet for the user to select the app of choice:


intent_chooser

PSchmiedmayer:
Nice, very cool @eldcn!

Issue #157, opened 2025-01-06
Health Survey Questionnaire Button Label


kylieo123:
### Problem

The button label that is intended to prompt the patient to fill out the survey is "Finish". This was confusing for me when I was playing with the app so i assume it will similarly be confusing to patients. 

### Solution

Change this label to "Start Questionnaire" to make it easier to understand. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PSchmiedmayer:
Nice, thank you for the report @kylieo123! 
CC: @pauljohanneskraft & @eldcn

vishnuravi:
This issue isn't relevant to this repository any longer. Closing.

PR #168, opened 2025-01-14
MessagesAction Update to also handle non-actionable messages without failing

pauljohanneskraft:
# MessagesAction Update to also handle non-actionable messages without failing

Previously, the app would handle a non-existing action as a failure, whereas it can simply be ignored and a message should then either be dismissible or will be dismissed automatically based on other user input.

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing

### DIFF
--- a/app/src/test/kotlin/edu/stanford/bdh/engagehf/bluetooth/HomeViewModelTest.kt
+++ b/app/src/test/kotlin/edu/stanford/bdh/engagehf/bluetooth/HomeViewModelTest.kt
@@ -333,6 +333,19 @@ class HomeViewModelTest {
     }
 
     @Test
+    fun `it should handle message dismiss action correctly`() {
+        // given
+        val isExpanded = false
+        createViewModel()
+
+        // when
+        viewModel.onAction(Action.MessageItemDismissed(messageId))
+
+        // then
+        coVerify { messagesHandler.dismiss(messageId) }
+    }
+
+    @Test
     fun `it should handle permission result action correctly`() = runTestUnconfined {
         // given
         val permissions = listOf("permission1")


app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/mapper/MessageActionMapper.kt
```diff
@@ -14,7 +14,7 @@ class MessageActionMapper @Inject constructor() {
     fun map(action: String?): Result<MessagesAction> {
         return runCatching {
```
eldcn, line 15: there is another error being thrown in the else case. I would propose changing the return type of this method to `MessageAction?` instead of result and update to `runCatching { // exisiting logic }.getOrNull()`. No need for NoAction imo as we can represent it with the null case

pauljohanneskraft:
I would like to discuss this further before merging, since there seem to be quite a few UI differences between messages on iOS and Android and I would rather discuss them before throwing out existing functionality here.

Examples:

- iOS has multi-line collapsing only after 3+ lines (by using a text button "Read more") and I'm not sure whether it is really intuitive there considering that if you tap on the collapsible message, the action is performed, so it is actually really hard to collapse and not perform the action instead. Android has that right chevron icon instead - We may want to think to just never collapse, right?

- iOS differentiates between dismissing and performing an action. Where you may dismiss a message on iOS without performing its action, you can only interact with a message using the "Finish" button on Android. Further, the "Finish" text is bound to the action type, which is not the case for Android.

- Due date is not displayed on iOS and may be removed on Android, since we do not set it on the server anymore (and if we do in some rare occasions, we do not need to visualize it).

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/Message.kt
```diff
@@ -15,7 +15,6 @@ data class Message(
     val description: String? = null,
     val action: String?,
     val isDismissible: Boolean = true,
-    val isLoading: Boolean = false,
```
eldcn, line 18: There is some logic in `HomeViewModel` that shows a loading indicator while a message is being processed, e.g. health summary pdf generation. I think we still need this

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/MessagesHandler.kt
```diff
@@ -30,6 +30,7 @@ class MessagesHandler @Inject constructor(
         val actionResult = messagesActionMapper.map(action = message.action)
         var failure = actionResult.exceptionOrNull()
         when (val messagesAction = actionResult.getOrNull()) {
+            is MessagesAction.NoAction -> Unit
```
eldcn, line 33: it should not be needed as it is covered in the else case (applies also in case you change the response to nullable instead of result)

Question: Should we complete the message below if we could not map any action for it?

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/Message.kt
pauljohanneskraft: I found this out now, yes - but I believe this should actually be in some wrapper type or something, since it has nothing to do with the model itself. Like this, I would assume that the server sends us this information, which is not the case.

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/MessagesHandler.kt
pauljohanneskraft: I'm not sure actually, would need to check iOS

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/Message.kt
eldcn: Totally agree, in general there should be distinct types for domain and ui layer, e.g. `Message` and `MessageUiModel` and currently the message is serving for both layers

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/MessagesHandler.kt
pauljohanneskraft: I checked: iOS only calls dismissMessage, if an action has been performed and the message has `isDismissible` set to true.

PSchmiedmayer:
@pauljohanneskraft Thank you for looking into this and thanks for the first feedback @eldcn!

Would be great to reduce the disparities between the different platforms.

> iOS has multi-line collapsing only after 3+ lines (by using a text button "Read more") and I'm not sure whether it is really intuitive there considering that if you tap on the collapsible message, the action is performed, so it is actually really hard to collapse and not perform the action instead. Android has that right chevron icon instead - We may want to think to just never collapse, right?
I think it could be fine to never collapse a message. We might still want to add a truncation mark after e.g. 10 lines to ensure that a long text or a bug doesn't break the UI?

> iOS differentiates between dismissing and performing an action. Where you may dismiss a message on iOS without performing its action, you can only interact with a message using the "Finish" button on Android. Further, the "Finish" text is bound to the action type, which is not the case for Android.
I like the way iOS is currently handling this. An X for any dismissible actions and a action button that is bound to the message type & let's you perform a message.

> Due date is not displayed on iOS and may be removed on Android, since we do not set it on the server anymore (and if we do in some rare occasions, we do not need to visualize it).
Agree, we can remove that.

> The isLoading property on Message shouldn't be in the model type but is rather UI-exclusive. Therefore, we could, for example, write a small wrapper type or something to add those UI-exclusive values.
Good point.

@eldcn & @pauljohanneskraft let me know if you need any additional context; happy to also sync on this on Tuesday if there are more open questions.

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/MessageAction.kt
```diff
@@ -0,0 +1,12 @@
...
+sealed interface MessageAction {
+    data object UnknownAction : MessageAction
+    data class VideoAction(val video: Video) : MessageAction
+    data object MedicationsAction : MessageAction
+    data object MeasurementsAction : MessageAction
+    data class QuestionnaireAction(val questionnaireId: String) : MessageAction
+    data object HealthSummaryAction : MessageAction
+}
+
+data class Video(val sectionId: String, val videoId: String)
```
eldcn, line 12: If this is used in context of message action only, I would be more in favor of simply declaring the properties in video action type instead of introducing this new type

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/models/UiState.kt
```diff
@@ -13,7 +16,26 @@ data class UiState(
...
     ),
     val missingPermissions: Set<String> = emptySet(),
-    val messages: List<Message> = emptyList(),
+    val messages: List<MessageUiModel> = emptyList(),
     val bluetooth: BluetoothUiState = BluetoothUiState.Idle(),
     val measurementDialog: MeasurementDialogUiState = MeasurementDialogUiState(),
 )
+
+data class MessageUiModel(
+    val message: Message,
```
eldcn, line 25: Wrapping the domain message in a ui model still violates why we discussed. Ideally the ui model should remain domain agnostic and only contain properties relevant for UI. It should be created in a UI mapper, e.g.

```kotlin
data class MessageUiModel(
    val id: String,
    val title: String,
    val description: String?,
    val dueDate: String?,
    val isExpanded: Boolean = false,
    val isLoading: Boolean = false,
    val icon: Int,
)

fun map(message: Message): MessageUiModel {
    return MessageUiModel(
        id = message.id,
        title = message.title,
        description = message.description,
        dueDate = message.dueDate?.format(DateTimeFormatter.ofPattern("MMM dd, yyyy hh:mm a")),
        isExpanded = false,
        isLoading = false,
        icon = when (message.action) {
            is MessageAction.MedicationsAction -> R.drawable.ic_medication
            is MessageAction.MeasurementsAction -> R.drawable.ic_vital_signs
            is MessageAction.QuestionnaireAction -> R.drawable.ic_assignment
            is MessageAction.VideoAction -> R.drawable.ic_visibility
            is MessageAction.UnknownAction -> R.drawable.ic_assignment
            is MessageAction.HealthSummaryAction -> R.drawable.ic_vital_signs
        },
    )
}
```

app/src/main/kotlin/edu/stanford/bdh/engagehf/messages/MessageAction.kt
pauljohanneskraft: sounds good to me honestly

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/models/UiState.kt
pauljohanneskraft: fine with this change as well, will do

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/mapper/MessageActionMapper.kt
pauljohanneskraft: Changed it accordingly

PSchmiedmayer:
@pauljohanneskraft Double-check that this resolves #157

app/src/main/kotlin/edu/stanford/bdh/engagehf/bluetooth/data/models/UiState.kt
pauljohanneskraft: mhmmm, okay, so I tried this, but usually when we interact with lower layers (e.g. when dismissing a message or handling an action based on a message), we would usually want the message actually... How would you usually solve this? Mapping back seems to not really solve any issues, since either data is lost or we have an implicit coupling (i.e. we would need to copy the other object property-by-property anyways)

After seeing this, I'm actually more in favor of wrapping the model type in the UI-Model, but open to discussions. @eldcn

eldcn: The lower layer does not need all the message props, but only the id, action and whether it is dismissable, similarly as you adapted the handler api to process firebase notifications. I see here two ways:

1. Store additionally `messageAction` and `isDismissable` in `MessageUiModel` since those are additionally needed for handling the message or
2. Cache the loaded messages list in a private property in the view model on every update before mapping. When having to handle a message due to `Action.MessageItemClicked`, you can query the cached list for the message id and hand over to the handler

eldcn (approved):

PR #183, opened 2025-02-23
Register, refresh and unregister device notification token

### LINKED ISSUE #179: Notification Token is not registered correctly
### Description

At the moment, the ENGAGE-HF Android app only calls the `registerDevice` function when a new notification token is available. Unfortunately, the function can only make use of that call, when the user is logged in when calling it and doesn't call it again on app startup.

### Reproduction

Open the app with a fresh install. Log into an account. Trigger a push notification (e.g. by having low weight observations and then adding a very high one). Notice that you do not receive a push notification, even though a new message is now visible on the home screen.

### Expected behavior

`registerDevice` should be called
- on startup of the app, after onboarding (even when it is not the first launch to also keep language/timeZone/appVersion/osVersion/etc up to date)
- when log in was successful
- when a new token is received
- when permission changed, so that now receiving permissions is allowed

`unregisterDevice` should be called
- right before a user logs out

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

eldcn:
# *Register, refresh and unregister device token*

- Fixes #179
- Persist last successfully registered token
- Introduce refreshToken mechanism and invoke on notification permission changes, app start up in case of registered users, and once the user logs in
- It is safe to invoke refresh token as it only submits, in case the current firebase message token is different from last submitted one, otherwise call gets ignored
- Add missing `unregisterDevice` function
- Invoke unregister device right before signout
- Sign out now changes to an async operation due to unregister call -> sign out item was adapted to an async item
- Extended and adapted existing tests

*Note*: I could not test notifications themselves, was however able to verify that token get submitted correctly on all events mentioned in #179. 

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing

### DIFF
--- a/core/notification/src/main/kotlin/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
+++ b/core/notification/src/main/kotlin/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
@@ -3,96 +3,120 @@ package edu.stanford.spezi.core.notification.fcm
 import android.content.Context
 import android.os.Build
 import com.google.firebase.functions.FirebaseFunctions
+import com.google.firebase.messaging.FirebaseMessaging
 import dagger.hilt.android.qualifiers.ApplicationContext
 import edu.stanford.spezi.core.coroutines.di.Dispatching
 import edu.stanford.spezi.core.logging.speziLogger
+import edu.stanford.spezi.modules.storage.di.Storage
+import edu.stanford.spezi.modules.storage.key.KeyValueStorage
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.tasks.await
 import java.util.Locale
 import java.util.TimeZone
 import javax.inject.Inject
+import javax.inject.Singleton
+
+interface DeviceRegistrationService {
+    fun registerDevice(token: String)
+    fun refreshDeviceToken()
+    suspend fun unregisterDevice()
+}
 
 /**
  * Service to register the device with the server.
  */
-internal class DeviceRegistrationService @Inject constructor(
+@Singleton
+internal class DeviceRegistrationServiceImpl @Inject constructor(
     @ApplicationContext private val context: Context,
     private val functions: FirebaseFunctions,
+    private val firebaseMessaging: FirebaseMessaging,
     @Dispatching.IO private val coroutineScope: CoroutineScope,
-) {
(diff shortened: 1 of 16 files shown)


PSchmiedmayer (approved):
Thank you @eldcn for working on this; great work! 🚀

core/notification/src/main/kotlin/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
```diff
@@ -3,96 +3,120 @@ package edu.stanford.spezi.core.notification.fcm
...
+                registerDevice(token = token)
+            }.onFailure { error ->
+                logger.e(error) { "Error while accessing firebase messaging token" }
+            }
         }
+    }
 
+    override suspend fun unregisterDevice() {
+        logger.i { "Received unregister device request" }
+        val storageToken = getStorageToken() ?: run {
```
pauljohanneskraft, line 90: If I see this correctly, then we wouldn't actually need to store the notification token, if there is a call (`firebaseMessaging.token`) to get the current token, right? Or is there harm in calling that function in some cases (e.g. triggering permission alerts or sth like that)?

The firebase function simply ignores, if a token is not registered already, so it wouldn't actually make a big difference, if we call it with an unregistered token (except environmentally, maybe...)

pauljohanneskraft (approved):
Just a quick question whether we actually even need to store the token - otherwise looks good!

core/notification/src/main/kotlin/edu/stanford/spezi/core/notification/fcm/DeviceRegistrationService.kt
eldcn: Nope, there is no harm to call that token method. it is just a blocking call and persisting locally is more lightweight compared to that call. furthermore, local storage is proposed mechanism to keep track token management of the device, e.g. to reduce number of unnecessary remote calls. you can read more in the [firebase docs](https://firebase.google.com/docs/cloud-messaging/manage-tokens) and in this firebase sample [here](https://github.com/firebase/quickstart-android/blob/0b77e46545ef38b040d140f360cfe3f75b54e15d/messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt#L131-L153)

pauljohanneskraft (approved):
Looks great! Thank you for making the adaptions! 😊

PR #187, opened 2025-03-07
Separate Spezi modules

pauljohanneskraft:
# Separate Spezi modules

This pull request separates Spezi code from the existing modules to form coherent packages that are similar in structure and content to the iOS equivalents. This is only the first step as this PR makes no changes to the actual code, but only structure. Separate PRs for each module will follow to similarize them to the iOS counterpart.

## :gear: Release Notes 

## :books: Documentation

## :white_check_mark: Testing


eldcn:
Thanks a lot for the work with restructuring. I checked out the project and reviewed locally, and will summarize all my feedback in this comment:

1. Since this repository will be the final state of Spezi project, I would have made use of this restructuring to come closer to that final state. I see we introduce now a `spezi` main directory and moved all spezi related modules there to separate them from the rest of the project. Moreover, this restructuring introduced a duplication in the namespace e.g. `edu.stanford.spezi.spezi.ui`, which we would have to adapt again when moving the apps out. I would have gone the other way around of putting all spezi modules top level and grouping the modules that are supposed to move out into dedicated parent directories. With the current setup, we would have to readapt the structure again, release the modules and then use them in other repositories. I finally propose this new (intermediate) structure:

```
.
└── SpeziKt/
    ├── engage-hf/
    │   ├── app
    │   ├── engage-specific-module-a
    │   ├── engage-specific-module-b
    │   └── and-so-on
    ├── intermediate/
    │   ├── list-of-remaning-modules-under-module
    │   └── list-of-remaining-modules-under-core
    ├── contact
    ├── core
    ├── credentialstorage
    ├── foundation
    ├── personalinfo
    ├── questionnaire
    ├── ui
    └── validation
```

2. I see you are pointing to `main`, but we still have [the open PR of heart beat app](https://github.com/StanfordSpezi/SpeziKt/pull/143) open, which would again require quite some adaption after these changes. Maybe I missed some details, but I thought this restructuring would have checked out the heart beat branch, to also minimize the needed adaptions after we merge this PR.
3. Next, I want to share some guidelines about the strucuture of android modules. A module encapsulates context related components that surve a specific purpose. Ideally, modules should be kept as small as possible (there is nothing wrong with a module containing a single file / component as part of its public api, see [okhttp-coroutines](https://github.com/square/okhttp/tree/master/okhttp-coroutines/src/main/kotlin/okhttp3/coroutines) or [androidx.compose.runtime:runtime-livedata](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/runtime/runtime-livedata/src/main/java/androidx/compose/runtime/livedata/)) to also respect the single responsibility principle and minimize the coupling with other modules. Keeping this in mind, the public api of a module is suggested to be put top level in the main package of the module itself. A further nesting of public api into additional packages should be avoided. In case there is such a need, e.g. `spezi:ui/helpers/theme`, this speaks for the theme package to be extracted into a dedicated module itself. In general I propose the following structure for a module:

```
.
└── SpeziKt/
    └── example-module/
        ├── internal
        ├── Component1.kt
        ├── Component2.kt
        └── Component3.kt
```

I can refer to [`androidx.compose.foundation:foundation-layout`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/commonMain/kotlin/androidx/compose/foundation/layout/) source code as a reference or a larger module such as [`androidx.compose.ui:ui`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/). As you can see all public api components are put top level, and the other packages only contain internal or private components within the module (high cohesion within the module). This approach not only achieves consistency and kind of forces us to evaluate the public api of a module, but makes it also clearer when looking at the imports of a file to easily derive where a dependency is coming from, e.g.

```kotlin
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.Column
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.foundation.layout.Row
```

Now considering our base namespace `edu.stanford.spezi`, and following this proposal, the next node in the namespace of a module would be the module itself, e.g. `edu.stanford.spezi.personalinfo`, which can result into these fully qualified names / imports:

```kotlin
import edu.stanford.spezi.personalinfo.NameFieldRow
import edu.stanford.spezi.personalinfo.NameTextField
import edu.stanford.spezi.personalinfo.PersonNameComponents
import edu.stanford.spezi.personalinfo.UserProfileComposable
```

4. When a module e.g. `spezi:validation` exposes components from another module e.g. `spezi:ui` in its public API, it must declare `spezi:ui` as an api dependency instead of implementation. This ensures that any project depending on `spezi:validation` automatically gets access to `spezi:ui` and avoids potential runtime errors.

```kotlin
dependencies {
    api(project(":spezi:ui"))
    // rest of dependencies
}
```

This is the equivalent of exported import in iOS. If not included as api dependency, this can lead to a `ClassNotFoundException` or `NoClassDefFoundError` at runtime when a consumer tries to access an exposed type from `spezi:ui` but does not have it in its own dependencies.

5. Restructured `spezi:core` module is not respecting single responsibility and I think we should avoid having such modules as are candidate modules to simply put new components that do not fit context wise to other modules. I propose to split into `spezi:logging` and `spezi:coroutines` or `spezi:concurrency`.

6. `spezi:credentialstorage`: The content under `/key` package is also a candidate to be put into a separate module which can then be used a dependency from `credentialstorage`. `KeyValueStorage` is a general purpose building block for creating persistence storages in general, and should not get defined under `credentialstorage`. E.g. see imports in `BLEPairedDevicesStorage`, even though we don't need `credentialstorage` functionalities, we are enforced to add it as a dependency.

7. We can think of grouping related modules into parent directories. E.g. I think such a parent directory can be `ui` which can include nested modules such as `ui:theme`, `ui:components`, `ui:testing` etc. Another proposal would be `persistence` which can include, `persistence:foundation` (KeyValueStorage), `persistence:localstorage`, `persistence:credentialstorage` etc.
8. `core:logging` can get deleted now (also from `settings.gradle`) and `core:design` can be party integrated into `ui`


Please let me know in case this is too much to be handled in one shot as part of this PR. I would be happy to assist to split the work together, as I have done these kind of restructurings quite extensively.

pauljohanneskraft:
Hey - thank you for your feedback:

1. I thought we had agreed on the current structure before? Sure, I will adapt it to all be on the outermost level, but I would have appreciated if we had decided on this earlier, since this will probably require some work that could have easily been avoided.

2. How would this be reasonably be possible? First of all, I assume that HeartBeat doesn't currently require many changes due to its limited dependencies to the Spezi ecosystem, but without merging HeartBeat to the main branch and then adapting this, it wouldn't have been possible to make changes to two branches within one PR. Maybe I'm misunderstanding, so could you please clarify?

3. Let's talk about this in more detail in another round of package-specific adaptions. I first of all tried to replicate the structure that iOS provides and in there, we also have a lot of different things crammed into the same module. While I agree that we shouldn't duplicate this problematic structure in general, I feel like this is way out of scope to be solved in this, since it may require more differentiation and thought, in addition to possible changes to the iOS codebase(s). Let's please postpone these kinds of discussions to a later stage.

4. Nice, thank you - didn't know about that!

5. See 3.

6. I agree - in general, I think we should think about keeping it out of the public API of that package or even integrating it directly into CredentialStorage instead. I couldn't do this without making changes to the ENGAGE-HF code and didn't want to include code changes in this PR to keep it reviewable. I would like to postpone this to when we do the adaptions to the CredentialStorage package.

7. That sounds great! Do you want me to do these adaptions now or can we please postpone them for now?

8. While these adaptions make sense for the codebase as it is now, I feel like we should make these integrations carefully and possibly integrate more towards the ENGAGE-HF codebase rather than moving it in the Spezi codebases. I tried to keep them as small as possible for now with the possibility to move more functionality into them at a later stage - would that be okay?

In general, I would love to get this merged rather soon than late and then do the rounds of adaptions, so that we can accelerate the development. While I appreciate the offer to help with the restructuring/split work, I feel like this would need to happen in separate PRs and possibly be at the stage of adapting individual modules to keep changes limited to a certain scope. This PR is only supposed to group code, so that we can then start adapting further.

eldcn (approved):
Looks great 🚀 Thanks for adaptions. I pushed a small fix for the test workflow. Feel free to merge once CI passes

Next, I would try to address modules:design and probably move some of the components there under :ui, e.g. /validated

PR #188, opened 2025-03-09
Delete and recreate encrypted storage file if key store corrupted

eldcn:
# *Encrypted storage*

I debugged the issue and it appears to happen if the key store used to encrypt the file becomes unavailable or corrupted. This can happen in several occasions such as os version updates or device data backup, and there is unfortunately no way to recover the data in such cases and we have to delete the previous file.

I was able to reproduce the issue in our integration tests: Start the emulator, and run the test suite of `storage` module. The tests produce shared prefs files generated with the key store of the last test session. Closing and relaunching the emulator, and rerunning the same test suite again triggers the crash because the key store signature between test sessions do not match. After the fix we handle this gracefully and recreate the file.

Screenshot 2025-03-09 at 11 26 55

Screenshot 2025-03-09 at 11 48 50

## :gear: Release Notes

## :books: Documentation

## :white_check_mark: Testing

### DIFF
--- a/modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorageFactory.kt
+++ b/modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorageFactory.kt
@@ -5,6 +5,7 @@ import android.content.SharedPreferences
 import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKey
 import dagger.hilt.android.qualifiers.ApplicationContext
+import edu.stanford.spezi.core.logging.speziLogger
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -20,6 +21,7 @@ internal class KeyValueStorageFactoryImpl @Inject constructor(
     private val storageFactory: KeyValueStorageImpl.Factory,
     @ApplicationContext private val context: Context,
 ) : KeyValueStorageFactory {
+    private val logger by speziLogger()
 
     private val masterKey: MasterKey by lazy {
         MasterKey.Builder(context)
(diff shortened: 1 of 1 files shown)


modules/storage/src/main/kotlin/edu/stanford/spezi/modules/storage/key/KeyValueStorageFactory.kt
```diff
@@ -36,23 +38,36 @@ internal class KeyValueStorageFactoryImpl @Inject constructor(
...
-                        EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
-                    )
+                KeyValueStorageType.UNENCRYPTED -> createUnencryptedStorage(fileName = fileName)
+
+                KeyValueStorageType.ENCRYPTED -> createEncryptedStorage(fileName = fileName) ?: run {
+                    logger.w { "First encrypted storage creation failed, deleting existing file and retrying..." }
+                    context.deleteSharedPreferences(fileName)
+                    createEncryptedStorage(fileName = fileName) ?: run {
+                        logger.w { "Second encrypted storage creation failed, returning a non encrypted file instead" }
+                        createUnencryptedStorage(fileName = fileName)
```
PSchmiedmayer, line 50: Wondering if that would be unexpected behavior; it might make sense to fail at that point rather than storing it in an unencrypted way?

eldcn: We should never land at this point, as after deleting the previous file, we will be able to create the fresh encrypted file. I just wanted to avoid throwing at this point as it would result into a runtime crash. As I said I am pretty confident that we won't land here, and can adapt to fail here instead.

PSchmiedmayer: Sounds good, if we should never get there crashing on purpose might be good; would allow us to collect crash reports and investigate why and how we could land here. Better than silently failing and storing the data in an unencrypted setup.

PSchmiedmayer (approved):
Looks great; thank you!

PR #189, opened 2025-03-12
Spezi Alignment

pauljohanneskraft:
# Spezi Alignment

This pull request improves on the current implementation to realize an aligned API with considerate deviations for a platform-native feel. It aims to align the API both more to Kotlin guidelines/best practices while providing feature parity for non-platform-specific code in the Spezi packages `foundation-*`, `core-*` and `ui-*`.

## :gear: Release Notes

## :books: Documentation

## :white_check_mark: Testing


ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/BasicNameFieldRow.kt
```diff
@@ -0,0 +1,141 @@
+package edu.stanford.spezi.ui.personalinfo
```
eldcn, line 1: A general comment about this module (can be ignored since we want to align with iOS)

I see the whole setup of wrapping and duplicating all params of base text field components to just achieve a row layout not ideal and flexible. We would have to go for the same replications in case we would need to support a column layout of description and text fields. And again the same replications if we would need to support RTL languages.

Additionally, the description content seems a bit redundant to me and deviating from material design while all the text fields support already the label setup. While the text fields will be usually put in a column / lazy column, with this approach we can't guarantee a consistent layout unless we set fix weights to the description and text fields. See the difference between the current approach vs using material design components directly

image


Also, the approach with manipulating the kproperty of the builder I see not needed and not common in android. Having to do this kind of builder setup for every component as we are doing for person name is a solution that does not scale. Compose is already providing the base setup and every use case can do that directly based on its design requirements. I can share this setup that I did in heart beat in a single file / component that support three styles: https://github.com/StanfordSpezi/SpeziKt/pull/143/files#diff-2010cd3c435eaece095576f781cc7cdccc913dbeb8078505fedf39bac6621a2bR33

image

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/UserProfile.kt
```diff
@@ -27,24 +27,46 @@ import androidx.compose.ui.unit.sp
...
 import edu.stanford.spezi.ui.Colors
 import edu.stanford.spezi.ui.ImageResource
-import edu.stanford.spezi.ui.ImageResourceComposable
 import edu.stanford.spezi.ui.SpeziTheme
 import edu.stanford.spezi.ui.StringResource
 import edu.stanford.spezi.ui.ThemePreviews
 import edu.stanford.spezi.ui.lighten
 import kotlin.math.min
 
+interface UserProfileImageLoader {
```
eldcn, line 36: Let's consider including [coil](https://github.com/coil-kt/coil) in a follow up. We can extend our Image resource with a case data class Remote(val url ...), and handle this internally to use coils [AsyncImage](https://github.com/coil-kt/coil/blob/main/coil-compose/src/commonMain/kotlin/coil3/compose/SingletonAsyncImage.kt#L46). It also supports api for fallbacks and place holders in case loading fails. And we dont need to create custom loaders for every use case then. For instance, we can apply it for Contact use case also

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/UserProfile.kt
```diff
@@ -27,24 +27,46 @@ import androidx.compose.ui.unit.sp
...
+    name: PersonNameComponents,
+    modifier: Modifier = Modifier,
+    imageLoader: UserProfileImageLoader?,
 ) {
     var size by remember { mutableStateOf(IntSize.Zero) }
     var loadedImage by remember(imageLoader) { mutableStateOf<ImageResource?>(null) }
 
-    LaunchedEffect(Unit) {
-        loadedImage = runCatching { imageLoader() }
+    LaunchedEffect(imageLoader) {
```
eldcn, line 68: Again, this setup would not be needed if we use coil, as it will make sure to launch coroutines and take care of caching internally

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/Validate.kt
```diff
@@ -37,18 +41,28 @@ fun Validate(
...
         ValidationEngineImpl(
             rules,
-            validationDebounce,
-            validationEngineConfiguration
+            configuration,
+            coroutineScope,
+            debounceDuration,
         )
     }
+    LaunchedEffect(configuration) {
```
eldcn, line 57: Launching coroutines in compose is expensive, and you are currently allocating and launching three. we can reduce into one LaunchedEffect(config, debounce, scope)

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/CompositionLocal.kt
```diff
@@ -0,0 +1,9 @@
+package edu.stanford.spezi.ui.validation
+
+import androidx.compose.runtime.compositionLocalOf
+
+val LocalValidationEngine =
```
eldcn, line 5: I see we provide this internally, is this expected to be configured from the callers? If so, we should not overwrite it would our own, if not we can mark it internal.

What about LocalValidationEngineConfiguration? I see no provide for it

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/ReceiveValidation.kt
```diff
@@ -4,19 +4,31 @@ import androidx.compose.runtime.Composable
...
+
+@Composable
+fun ReceiveValidation(
+    onChange: (ValidationContext) -> Unit,
+    content: @Composable () -> Unit,
 ) {
     // This is not remembered on purpose, since we are re-evaluating the validation here.
-    val entries = CapturedValidationStateEntries()
-    CompositionLocalProvider(LocalCapturedValidationStateEntries provides entries) {
+    val entries = mutableListOf<CapturedValidationState>()
```
eldcn, line 26: this will be reset to empty in case every recomposition, we should use remembered mutable state instead

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/ValidationContext.kt
```diff
@@ -1,49 +1,50 @@
 package edu.stanford.spezi.ui.validation
 
-data class ValidationContext internal constructor(
+data class ValidationContext(
```
eldcn, line 3: Should we review internal / public components as part of this PR?

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/ValidationRule.kt
```diff
@@ -4,32 +4,39 @@ import edu.stanford.spezi.foundation.UUID
...
     val rule: (String) -> Boolean,
     val message: StringResource,
     val effect: CascadingValidationEffect = CascadingValidationEffect.CONTINUE,
 ) {
+    companion object
+
+    // Properties
+
+    @Suppress("detekt:VariableMinLength")
+    internal val id: UUID = UUID()
```
eldcn, line 21: feel free to adapt detekt config file to exclude id from the variable min length rule

ui-validation/src/test/kotlin/edu/stanford/spezi/ui/validation/SpeziValidationTest.kt
```diff
@@ -1,13 +1,17 @@
...
 import org.junit.Test
 
 class SpeziValidationTest {
 
     @Test
     fun testValidationDebounce() {
-        val engine = ValidationEngineImpl(rules = listOf(ValidationRule.nonEmpty))
+        val engine = ValidationEngineImpl(
+            rules = listOf(ValidationRule.nonEmpty),
+            coroutineScope = GlobalScope
```
eldcn, line 13: We should use test scopes for tests, consider SpeziTestScope or TestScope directly

ui/src/main/kotlin/edu/stanford/spezi/ui/ComposableContent.kt
```diff
@@ -0,0 +1,9 @@
+package edu.stanford.spezi.ui
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+interface ComposableContent {
```
eldcn, line 6: Consider adding 

```kotlin
@Composable
    fun Content() {
        Content(Modifier)
    }
```

in case of default modifiers, as default values are not supported in composable interface methods

ui/src/main/kotlin/edu/stanford/spezi/ui/ImageResource.kt
```diff
@@ -30,4 +39,50 @@ sealed interface ImageResource {
...
         override val identifier = UUID().toString()
     }
+
+    @Composable
+    override fun Content(modifier: Modifier) {
+        Content(modifier = modifier, tint = Colors.primary)
+    }
+
+    @Composable
+    fun Content(modifier: Modifier, tint: Color) {
```
eldcn, line 49: Consider introducing this typealias https://github.com/StanfordSpezi/SpeziKt/pull/143/files#diff-5e1b1e9082809c2ba95dc5c5fbdf0b7d44d4167b9c9e8ea87dd50e1076a21290R19

See doc comments there, you can have than `tint: ComposeValue<Color> get() { Colors.primary } as an interface requirement with default value, this way we can stick to the base ComposableContent interface, and can extend the type instead of introducing new methods for every config option

contact/src/main/kotlin/edu/stanford/spezi/contact/ContactContent.kt
```diff
@@ -38,17 +37,17 @@ import java.util.Locale
...
+ * @sample ContactContentPreview
  *
  * @see Contact
  * @see ContactOptionCard
  * @see AddressCard
  */
 @OptIn(ExperimentalLayoutApi::class)
 @Composable
-fun ContactComposable(contact: Contact, modifier: Modifier = Modifier) {
+fun Contact.Content(modifier: Modifier = Modifier) {
```
eldcn, line 50: Anything against conforming Contact directly to ComposableContent? It is indeed a ui layer model as it already has ui properties such as image resource and string resources

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/BasicNameTextField.kt
```diff
@@ -0,0 +1,126 @@
...
+    onTextLayout: (TextLayoutResult) -> Unit = {},
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    cursorBrush: Brush = SolidColor(Color.Black),
+    decorationBox: @Composable (innerTextField: @Composable () -> Unit) -> Unit =
+        @Composable { innerTextField -> innerTextField() },
+) {
+    BasicTextField(
+        value = value,
+        onValueChange = onValueChange,
+        modifier = modifier.testIdentifier(
```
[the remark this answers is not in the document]
pauljohanneskraft: Yes, let's remove both all NameTextField and NameFieldRow variations, as they do not provide a lot of functionality anyways.

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/KeyboardOptions.kt
```diff
@@ -0,0 +1,15 @@
+package edu.stanford.spezi.ui.personalinfo
+
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.ui.text.input.KeyboardCapitalization
+import androidx.compose.ui.text.input.KeyboardType
+
+val KeyboardOptions.Companion.NameDefault
```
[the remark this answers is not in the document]
pauljohanneskraft: We may want to keep exposing this as it is actually the only real override for the NameTextField - so replacing a NameTextField from Swift to Kotlin would be using a regular TextField and setting this one value.

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/BasicNameFieldRow.kt
pauljohanneskraft: DescriptionGridRow, NameTextField and NameFieldRow can be removed from this package.

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/UserProfile.kt
pauljohanneskraft: Yes, we can provide an AsyncImageResource in the ui-package and make use of it here - I would actually keep it separate from ImageResource though to make a clear separation between cases where we allow for asynchronous loading and others where async loading would need to be handled separately and we actually expect an image to be available immediately.

ui-personalinfo/src/main/kotlin/edu/stanford/spezi/ui/personalinfo/UserProfile.kt
pauljohanneskraft: See above

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/Validate.kt
pauljohanneskraft: I'm fine with combining them

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/CompositionLocal.kt
pauljohanneskraft: We need to expose this as it allows to write custom components similar to ValidatedTextField

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/ReceiveValidation.kt
pauljohanneskraft: Yes, this is on purpose. The individual entries are remembered, but we need to recompute this list on every run.

ui-validation/src/main/kotlin/edu/stanford/spezi/ui/validation/ValidationRule.kt
pauljohanneskraft: Fine with that

ui-validation/src/test/kotlin/edu/stanford/spezi/ui/validation/SpeziValidationTest.kt
pauljohanneskraft: Sure, didn't know about that - sounds good!

ui/src/main/kotlin/edu/stanford/spezi/ui/ImageResource.kt
pauljohanneskraft: I imagine ImageResource to not really make assumptions about the appearance in UI, but more like an object to know how to get some image data... In the UserProfile for example, we want to use the image as a mask and then apply a custom tint that isn't provided by an app developer but instead set by spezi:ui-personalinfo.

We may want to think about having two interfaces `ComposableContent` and `ConfigurableComposableContent<Configuration>` maybe, where one allows to have a generic constraint with a given configuration to set some values during composition. In this case, we could then create a data class `ImageResourceConfiguration` with the tint as one of the properties - or sth like that.

contact/src/main/kotlin/edu/stanford/spezi/contact/ContactContent.kt
pauljohanneskraft: Nope, just that we would then need to put it in the file of the Contact itself and wouldn't have separated UI code from Compose stuff, but actually I'm fine with moving it, if you want.

eldcn (approved):

modules/education/src/main/java/edu/stanford/spezi/modules/education/videos/component/ExpandableVideoSection.kt
```diff
@@ -156,13 +154,13 @@ private fun VideoItem(video: Video, onVideoClick: () -> Unit) {
                 .padding(Spacings.small)
                 .fillMaxWidth()
         ) {
-            AsyncImageResourceComposable(
-                imageResource = AsyncImageResource.Remote(url = video.thumbnailUrl, StringResource("Video thumbnail")),
+            AsyncImageResource(
```
pauljohanneskraft, line 157: Don't you need to call Content on the resource then? I'd assume that the resource itself isn't a composable?

ui/src/main/kotlin/edu/stanford/spezi/ui/AsyncButton.kt
```diff
@@ -19,13 +19,10 @@ import androidx.compose.ui.Modifier
...
-import edu.stanford.spezi.ui.Sizes
-import edu.stanford.spezi.ui.SpeziTheme
-import edu.stanford.spezi.ui.ThemePreviews
-import edu.stanford.spezi.ui.testing.testIdentifier
+import edu.stanford.spezi.ui.theme.Colors
+import edu.stanford.spezi.ui.theme.Sizes
+import edu.stanford.spezi.ui.theme.SpeziTheme
+import edu.stanford.spezi.ui.theme.ThemePreviews
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
```
pauljohanneskraft, line 27: I was hesitant so far to copy over this component, since I'd usually assume that `AsyncTextButton` is the `TextButton` with async functionality, which isn't exactly the case, since it provides some other defaults as the `TextButton` has

In general, we may use a similar approach here as we planned to use for the TextFields (i.e. a shared interface with some styling parameter to allow for easier wrapping across all types of textfields/buttons), so happy to hear your thoughts!

ui/src/main/kotlin/edu/stanford/spezi/ui/AsyncImageResource.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.ContentScale
+import coil.compose.SubcomposeAsyncImage
+
+/**
+ * Composable function to display an image from a remote url
+ */
+@Composable
+fun AsyncImageResource(
```
pauljohanneskraft, line 16: Okay - not I see it here, that it is in fact a Composable - I would usually expect it to not be one though 😅 similar as to the ImageResource, I'd prefer an approach where a viewModel or even some service could define a resource and we could then specify loading/error ui when we actually define the UI around that AsyncImageResource, right?

pauljohanneskraft (commented):
4 discussion points, nothing blocking - let me know what you think! 😊

PR #198, opened 2025-04-25
Add Spezi core module dependencies setup

eldcn:
# *Add Spezi core module dependencies setup*

- Adds singleton scoped based DI DSL for Spezi Modules
- Introduce `SpeziApplication` interface with `Configuration` requirement
- Client apps are expected to only conform their Android application class to `SpeziApplication`, and the library will take care internally to resolve the dependencies. This is happening via a `ContentProvider`, which is an android component that OS creates before calling `Application#onCreate()` and gives access to application context
- For this iteration: accessing of dependencies is not restricted to any specific components but it's rather being made available globally similarly to iOS
- Unit tested via a test application

API example:

```kotlin
class MyApplication : Application(), SpeziApplication {
    private val customConfiguration = Configuration {
        module { AudioModule() }
        module { Preprocessor(module = dependency()) }
        module {
            CoughModule(
                audioModule = dependency(),
                preprocessor = dependency()
            )
        }
    }

    override val configuration: Configuration = Configuration {
        include(configuration = customConfiguration)
        module<Onboarding> { OnboardingImpl() }
        module { Module1(name = "Module 1") }
        module { Module2(age = "Module 2") }
        module("my-module-x") { ModuleX() }
        module {
            ModuleAlternative(
                module1 = dependency<Module1>(),
                module2 = dependency<Module2>()
            )
        }
        module(identifier = "cough-module-with-identifier") { dependency<CoughModule>() }
    }
}

class ExemplaryClassAccessingDependencies {
    val onboarding by dependency<Onboarding>()
    val module1 by dependency<Module1>()
    val module2 by dependency<Module2>()
    val moduleX by optionalDependency<ModuleX>("my-module-x")
    val coughModuleWithKey by dependency<CoughModule>(identifier = "cough-module-with-identifier")
}
```

## :gear: Release Notes

## :books: Documentation

## :white_check_mark: Testing

### DIFF
--- a/core/src/main/kotlin/edu/stanford/spezi/core/Configuration.kt
+++ b/core/src/main/kotlin/edu/stanford/spezi/core/Configuration.kt
@@ -1,0 +1,58 @@
+package edu.stanford.spezi.core
+
+import edu.stanford.spezi.core.internal.ModuleRegistry
+
+/**
+ * A configuration for the Spezi framework, which allows you to define and register modules and their dependencies.
+ */
+sealed interface Configuration {
+
+    companion object {
+        /**
+         * Creates a new [Configuration] instance using the provided configuration block.
+         *
+         * This function allows you to define and register modules and their dependencies in a declarative way in your application.
+         *
+         * Example usage:
+         *
+         * ```kotlin
+         *
+         * class MyApplication : Application(), SpeziApplication {
+         *     override val configuration: Configuration = Configuration {
+         *          module { AudioModule() }
+         *          module<Onboarding> { OnboardingImpl() }
+         *          module(identifier = "alternative-onboarding") { AlternativeOnboarding() }
+         *          include(configuration = externalConfiguration)
+         *     }
+         * }
+         * ```
+         *
+         * @param scope The configuration block to configure the [Configuration].
+         * @return A new [Configuration] instance.
+         */
+        operator fun invoke(
+            scope: ConfigurationBuilder.() -> Unit,
+        ): Configuration = ConfigurationBuilder().apply(scope).build()
+    }
+}
+
+/**
+ * Combines two [Configuration] instances into a new one, merging their modules and factories.
+ */
+operator fun Configuration.plus(other: Configuration): Configuration {
+    val registry = ModuleRegistry()
+    registry.register(configuration = this)
+    registry.register(configuration = other)
+    return ConfigurationImpl(registry = registry)
+}
+
+/**
+ * A internal implementation of the [Configuration] interface.
+ *
+ * @param registry The [ModuleRegistry] that holds the registered modules and their dependencies.
+ */
+@PublishedApi
+internal data class ConfigurationImpl internal constructor(
+    @PublishedApi
+    internal val registry: ModuleRegistry,
+) : Configuration
--- a/core/src/main/kotlin/edu/stanford/spezi/core/ConfigurationBuilder.kt
+++ b/core/src/main/kotlin/edu/stanford/spezi/core/ConfigurationBuilder.kt
@@ -1,0 +1,45 @@
+package edu.stanford.spezi.core
+
+import edu.stanford.spezi.core.internal.ModuleKey
+import edu.stanford.spezi.core.internal.ModuleRegistry
+
+/**
+ * Builder for creating a [Configuration] for a [SpeziApplication].
+ */
+class ConfigurationBuilder internal constructor() {
+    @PublishedApi
+    internal val registry = ModuleRegistry()
+
+    /**
+     * Registers a module factory.
+     *
+     * @param identifier An optional identifier key associated with the instance built via [factory].
+     * The same identifier can be used to register multiple instances of the same module type and can then
+     * be used to retrieve the module via [dependency] or [optionalDependency].
+     * @param factory Factory building scope executed in the completely built [DependenciesGraph] that returns an instance of the module.
+     */
+    inline fun <reified M : Module> module(
+        identifier: String? = null,
+        noinline factory: DependenciesGraph.() -> M,
+    ) {
+        registry.register(key = ModuleKey(identifier = identifier), factory = factory)
+    }
+
+    /**
+     * Registers a an existing configuration and its associated modules and factories to self.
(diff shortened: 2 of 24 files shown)


core/src/main/kotlin/edu/stanford/spezi/core/Module.kt
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Base interface that all Spezi modules must implement in order to get provided in the [SpeziApplication] modules dependency graph.
+ */
+interface Module {
+
+    /**
+     * Optional configuration method that can be overridden by the module to perform any setup or initialization. This method is called
+     * by the Spezi framework after all module has been registered in the dependency graph.
+     */
+    fun configure() {}
```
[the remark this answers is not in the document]
eldcn: yep, it is already the case, the empty function body does that `{}`

core/src/main/kotlin/edu/stanford/spezi/core/DependenciesGraph.kt
```diff
@@ -0,0 +1,111 @@
...
+import edu.stanford.spezi.core.internal.ModuleRegistry
+import edu.stanford.spezi.core.internal.speziCoreLogger
+import edu.stanford.spezi.foundation.simpleTypeName
+
+/**
+ * A graph of dependencies built at app start up via the configuration of the [SpeziApplication].
+ *
+ * @param registry The [ModuleRegistry] containing the registered modules and factories.
+ */
+class DependenciesGraph internal constructor(
```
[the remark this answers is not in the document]
eldcn: yep, the instance itself and two of the methods need to be public, `dependency` and `optionalDependency`. this is the scope passed to the module factory in order to read the previously registered modules `module { // here we can access `dependency` and `optionalDependency` }`

core/src/main/kotlin/edu/stanford/spezi/core/Configuration.kt
PSchmiedmayer, line 1: Thank you for adding all the documentation here in in the other places!

core/src/main/kotlin/edu/stanford/spezi/core/Dependency.kt
```diff
@@ -0,0 +1,47 @@
...
+import edu.stanford.spezi.core.internal.Spezi
+
+/**
+ * Lazy delegate to retrieve an optional module dependency from the [SpeziApplication] dependency graph.
+ *
+ * Example usage:
+ *
+ * ```kotlin
+ *
+ * class MyComponent {
```
PSchmiedmayer, line 13: Wondering if we should make the examples use `Module`s themselves? Technically not required but might be good to incentivize this; we explicitly restrict the usage of `@Dependency` on iOS to within `Modules`. But I think we can't easily do this on Android + don't really have a way to detect that this is the case or that it is injected somewhere in the view hierarchy.

core/src/test/kotlin/edu/stanford/spezi/core/internal/SpeziTests.kt
PSchmiedmayer, line 1: Thank you for adding all the tests here!

PSchmiedmayer (approved):
Impressive work; thank you @eldcn!

It would be amazing if we can take the example for them PR description (and maybe some larger context about the framework) and add that to the README. We can make a disclaimer that we are not at the same state as iOS at this point, but a nice entry way to explain the core functionality and what we want to do similar to the iOS version would be great next step.

core/src/main/kotlin/edu/stanford/spezi/core/Dependency.kt
eldcn: It is possible to restrict accessing of dependencies to certain components. Currently both `dependency` and `optionalDependency` methods are declared as public methods, but we could for instance declare those as extension functions of a certain type, e.g. in order to restrict only within the modules, we could do something like this:

```kotlin

inline fun <reified M : Module> Module.dependency(identifier: String? = null): Lazy<M> = lazy {
    val key = ModuleKey<M>(identifier)
    Spezi.requireGraph().optionalDependency(key)
        ?: speziError("$key not found. Please make sure to register via in the configuration block of your app component")
}

class Module1 : Module

class Module2 : Module {
    val someDependency by dependency<Module1>()
}

class NonModule {
    val someDependency by dependency<Module1>() // This will not compile
}
```

What we unfortunately can't do in kotlin, is declaring for instance a `DependencyAccessor` marker interface, and only make accessing of dependencies for the types conforming to that interface. While this would work for new types, it does not work for types declared in other dependencies, e.g. an android Fragment, ViewModel or Activity. The reason is that kotlin does not support interface conformance extension, e.g. extending Fragment to conform to our custom DependencyAccessor interface.

Let's discuss as follow up in which components do we want to access the modules. I would imagine within Module interface, Composable, ViewModel and whenever an android `Context` is available

core/src/main/kotlin/edu/stanford/spezi/core/ConfigurationBuilder.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Registers a module factory.
+     *
+     * @param identifier An optional identifier key associated with the instance built via [factory].
+     * The same identifier can be used to register multiple instances of the same module type and can then
+     * be used to retrieve the module via [dependency] or [optionalDependency].
+     * @param factory Factory building scope executed in the completely built [DependenciesGraph] that returns an instance of the module.
+     */
+    inline fun <reified M : Module> module(
```
RealLast, line 21: Wondering if maybe "register" or "loader" could be a more intuitive name for this function?
We could do:

``` kotlin
load { MyModule() }
```

or

``` kotlin
register { MyModule() }
```

PSchmiedmayer:
Thank you for the review @RealLast!

I think the functional to implicit create modules that are default initializable should ideally be implicit generated without listing them in the configuration. We should have the same base functionality on iOS and Android to enable this behavior and ideally the mental burden on the developer to translate elements between platforms. This might not be required for this PR; we can see how we can address these improvements step-by-step?

PR #202, opened 2025-05-07
Make sure no two register requests are performed at the same time

pauljohanneskraft:
# Make sure no two register requests are performed at the same time

We are currently experiencing multiple Firebase Auth users with the same email address and different userIds. This is usually avoided, since Firebase Auth checks, if there is an existing user with the same email address already (since this is also the only identifier when logging in, you otherwise get corrupted user accounts). Apparently, Firebase doesn't guard us from race conditions (i.e. creating multiple accounts with the same email address at the same time). On iOS, we already shield this from happening by disabling the register button while the registration is in process.

## :gear: Release Notes

## :books: Documentation

## :white_check_mark: Testing


eldcn (approved):
LGTM 🚀 as discussed the async button adaption would be just a nice to have addition :)

PR #203, opened 2025-05-11
Default Module initialization

eldcn:
# *Default Module initialization*

Additional attempts to create a module via reflection when a dependency not registered is requested. Order of attempts:
1. Check whether the module type provides an empty constructor
2. Check whether the module type provides a `constructor(context: Context)` constructor
3. Check whether the companion object of the module conforms to `DefaultInitializer<M : Modules>`
4. Else throw

Applied the default initializer approach in LocalStorage.

## :gear: Release Notes

## :books: Documentation

## :white_check_mark: Testing


core/src/test/kotlin/edu/stanford/spezi/core/internal/SpeziTests.kt
PSchmiedmayer, line 1: Nice to see the examples, looks great!

PSchmiedmayer (approved):
Nice setup, looks great! 🚀

core/src/main/kotlin/edu/stanford/spezi/core/DependenciesGraph.kt
```diff
@@ -39,8 +41,40 @@ class DependenciesGraph internal constructor(
...
+        return if (instance != null) {
+            registry.modules[typeKey] = instance
+            logger.w { "Instantiated module $typeKey manually via fallback mechanism. Consider registering it explicitly." }
+            instance
+        } else {
+            speziError(
+                message = "$typeKey not found. Please make sure to register via in the configuration block of your app component",
+                cause = result.exceptionOrNull()
+            )
+        }
```
pauljohanneskraft, line 77: What about constructors with default parameters, do they count as having less parameters, i.e. does the compiler just generate multiple versions of the same constructor with a separate set of parameters? 

I could for example imagine something like that, please let me know, if it doesn't make sense or I'm misinterpreting something:

```kotlin
class CredentialStorage(
    val context: Context, 
    val localStorage: LocalStorage = dependency<LocalStorage>()
) {}
```

This should still be automatically injected when needed, right? I've seen the approach with the create-function, but I'm just wondering, if we can create a more developer-friendly experience here.

pauljohanneskraft (approved):
Looks good to me in terms of the requirements specified in the pull request description. I have one small remark on how we might be able to improve developer experience, let me know, if that's even possible and/or reasonably easy to implement!

core/src/main/kotlin/edu/stanford/spezi/core/DependenciesGraph.kt
pauljohanneskraft: Okay - I'm fine with the annotation workaround in these cases, so all good - thank you!

PR #214, opened 2026-04-27
View model support for Spezi DI

eldcn:
# *View model support for Spezi DI*

Removes remaining Hilt usages across modules and replaces them with the Spezi core DI. Also adds missing ViewModel support to fully close the feature gap with Hilt.

- **Drop Hilt** in favor of our custom Spezi DI: removed `@HiltAndroidApp`, `@AndroidEntryPoint`, Hilt test runners, and all related Hilt modules/component bindings across affected modules
- **ViewModel DI** – added `speziViewModel<VM>()` as a first-class composable, the direct analogue of Hilt's `hiltViewModel<VM>()`, with matching default parameter signatures (`viewModelStoreOwner`, `key`, `factory`, `extras`)
- **Fragment & Activity support** – companion retrievers for `Fragment` and `ComponentActivity` mirror the composable API
- **`viewModels { }` DSL** – factories are registered inside the module graph via `viewModels { viewModel { MyVm(dependency()) } }`, keeping the same declarative style as the rest of the DI
- **Singleton & factory support** – extended the DI graph with `singleton { }` and `factory { }` registrations alongside module-based bindings, so any arbitrary type can be provided and resolved via `dependency<T>()` / `optionalDependency<T>()` without wrapping it in a module
- Further clean ups in module dependencies and convention plugins
- Unit tests
- Code example:

```kotlin
// Configuration block
Configuration(standard = AppStandard) {
    module { AppModule() }
    viewModels {
        viewModel { HomeViewModel(dependency()) }
        viewModel { DetailViewModel(dependency(), savedStateHandle()) }
    }

    // Singleton
    singleton { AnalyticsService(apiKey = "key") }

    // Factory – new instance on every resolution
    factory { RequestLogger(dependency<AnalyticsService>()) }
}

// Retrieval in a Composable – mirrors hiltViewModel<VM>()
@Composable
fun HomeScreen() {
    val viewModel = speziViewModel<HomeViewModel>()
    // ...
}
```

## :gear: Release Notes

## :books: Documentation

## :white_check_mark: Testing


core/src/main/kotlin/edu/stanford/spezi/core/internal/DependencyRegistry.kt
```diff
@@ -0,0 +1,93 @@
...
+        factories[key]?.let { factory ->
+            return graph.keyResolvingScope(key) {
+                val instance = graph.factory() as T
+                instances[key] = instance
+                factories.remove(key)
+                instance
+            }
+        }
+        return transientFactories[key]?.let { it(graph) as T }
+    }
```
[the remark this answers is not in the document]
eldcn: The race window only exists between graph construction and the end of resolveFactories(), which runs synchronously on a single thread on Android startup. The condition is theoretical and not reachable in the current design. The fix is not needed and would introduce unnecessary locking complexity with a deadlock risk.

core-viewmodel/src/test/kotlin/edu/stanford/spezi/core/viewmodel/ViewModelsTests.kt
```diff
@@ -0,0 +1,156 @@
+package edu.stanford.spezi.core.viewmodel
```
pauljohanneskraft, line 1: I'm thinking about whether this would kind of make sense to be put under edu.stanford.spezi.ui.viewmodel instead - what do you think?

sample-app/src/main/kotlin/edu/stanford/spezi/sample/app/SampleApplication.kt
```diff
@@ -11,20 +10,34 @@ import edu.stanford.spezi.core.Configuration
...
 class SampleApplication : Application(), SpeziApplication, HealthConstraint {
     private val logger by speziLogger()
 
     override val configuration: Configuration = Configuration(standard = this) {
-        module {
-            Navigator(concurrency = dependency())
+        singleton { Navigator() }
+
+        viewModels {
+            viewModel {
```
pauljohanneskraft, line 30: Would it possibly make sense to not have this one indirection and instead write the `viewModel` in the outer scope directly? I get the idea of being able to group them, but that can easily also be done by the developer of an application, not necessarily a need to force it on them, right?

Unless there is a specific reason of course for the viewmodels to be aggregated into one object before.

pauljohanneskraft (approved):
Two small things I noticed - otherwise looks great! Thank you for extracting Hilt and generally making sure we move all the Spezi modules to the new DI system!

